PR #198, opened 2014-03-25
Change composer dependency to "require-dev"

yourwebmaker:
As this tool is only used to dev purposes, I think the composer dependency must be set only for dev env.


rianorie:
I agree with @natronite, I use the migrations in my production environment as well

PR #244, opened 2014-09-05
Fixed issue with getSource() when forcing model generation

natronite:
The dev tools somehow added a second getSource() when generating models if run with the --force parameter. This is no longer the case.


rianorie:
Great effort! However.. there's too many changes in a single PR, please separate them and verify the getSource fix please, it didn't work for me.

PR #248, opened 2014-09-24
Fix on webtools for generate model

RaphaelNeumann:
Litter fix to correct include 'schema' in model generate


sergeyklay:
Thanks

PR #266, opened 2014-11-17
Changed path to migration directory

vitaliykoziy:
APP_DIRECTORY = this is constant which we set in webtools.php define("APP_DIRECTORY", __DIR__ . DIRECTORY_SEPARATOR . '..' . DIRECTORY_SEPARATOR); 
I should use this because I don't have app/ foldaer.


rianorie:
related to #265

sergeyklay:
For migration dir need use command line option or implement config reading

PR #268, opened 2014-11-17
was error for PostgreSql

vitaliykoziy:
Notice: Undefined index  schema in \Phalcon\Builder\Model.php on line 341

### DIFF
--- a/scripts/Phalcon/Builder/Model.php
+++ b/scripts/Phalcon/Builder/Model.php
@@ -338,7 +338,7 @@ class Model extends Component
         } elseif ($adapter == 'Postgresql') {
             $schema = 'public';
             $initialize[] = sprintf(
-                $templateThis, 'setSchema', '"' . $this->_options['schema'] . '"'
+                $templateThis, 'setSchema', '"' . $schema . '"'
             );
         } else {
             $schema = $config->database->dbname;


rianorie:
@vitaliykoziy are you familiar with Postgres at all? Can you explain why we're defaulting to the `public` schema in Postgres?

vitaliykoziy:
Hi,  @rianorie .
Of course, that the scheme may be different, you can call it whatever you like.
But, this fix is not for scheme name.

```php

        if (isset($this->_options['schema'])) {
            if ($this->_options['schema'] != $config->database->dbname) {
                $initialize[] = sprintf(
                    $templateThis, 'setSchema', '"' . $this->_options['schema'] . '"'
                );
            }
            $schema = $this->_options['schema'];
        } elseif ($adapter == 'Postgresql') {
            $schema = 'public';
            $initialize[] = sprintf(
                $templateThis, 'setSchema', '"' . $schema . '"'// this is my fix
                $templateThis, 'setSchema', '"' . $this->_options['schema'] . '"'// this is old code
            );
        }

```

And my question. How can You call variable ($this->_options['schema'] ) which not isset  for  block `elseif `?

vitaliykoziy:
And, yes, I familiar with Postgres.

rianorie:
```
$schema = 'public';
$initialize[] = sprintf($templateThis, 'setSchema', '"' . $schema . '"');
```
will end up as:
```
this->setSchema("public");
```
In MySQL that would mean we're setting the database name as `public`. Is that the same in Postgres?

I'm thinking, looking back, that maybe this would've been better:

```
        if (isset($this->_options['schema'])) {
            if ($this->_options['schema'] != $config->database->dbname) {
                $initialize[] = sprintf(
                    $templateThis, 'setSchema', '"' . $this->_options['schema'] . '"'
                );
            }
            $schema = $this->_options['schema'];
        } elseif ($adapter == 'Postgresql') {
            if ($this->_options['schema'] != $config->database->dbname) {  // added
                $initialize[] = sprintf(
                    $templateThis, 'setSchema', '"' . $this->_options['schema'] . '"'  // original line
                );
            } // added
        }
```

vitaliykoziy:
when You will create database, it will has `public` scheme by default. But You can create new scheme for this database.

but, `$templateThis, 'setSchema', '"' . $this->_options['schema'] . '"' ` variable `$this->_options['schema']` not isset, do you agree with me, about this ?

rianorie:
Ok, got it. So schema != database in Postgres, thanks for clearing that up.

How's this (I grabbed a few more lines from the file):

```php
        $initialize = array();
        if (isset($this->_options['schema'])) {
            if ($adapter == 'Postgresql') {       
                $initialize[] = sprintf(
                    $templateThis, 'setSchema', '"' . $this->_options['schema'] . '"'
                 );
                 $schema = $this->_options['schema'];

            } else { // kind of assuming Mysql here, but the whole model does that.
                if ($this->_options['schema'] != $config->database->dbname) {
                    $initialize[] = sprintf(
                        $templateThis, 'setSchema', '"' . $this->_options['schema'] . '"'
                     );
                }
                $schema = $this->_options['schema'];
            }
        } else {
            $schema = $config->database->dbname;
        }
```

For the record, I realize there's room for improvement. For now I'm just trying to get the PRs merged and then I'll do a rewrite of the entire repository.

vitaliykoziy:
And so interesting for me, maybe You can help me, i hope.
I think that in phalcon, we can not connect to differents schemas of database. We can connect only to public schema, I think so.
How you think, is it true ? :)

rianorie:
I can't say for sure.. I have no experience with Postgres, Looking for that shouldn't go in this PR though :) Perhaps ping me on freenode (IRC) in the phalconphp channel, or throw a post on the phalcon forums. Likely Postgres other users have been experiencing this and if it's true, it's something we need to get fixed.

vitaliykoziy:
and about your new lines of code...
In prev was  
```php
if (isset($this->_options['schema'])) { 
//code
}
elseif ($adapter == 'Postgresql') {
//code
}
```
and as You know, this `isset($this->_options['schema'])` return `FALSE` when i use postgreSql.

I about that, in your new code 
```php
if (isset($this->_options['schema'])) {
            if ($adapter == 'Postgresql') {       
                $initialize[] = sprintf(
                    $templateThis, 'setSchema', '"' . $this->_options['schema'] . '"'
                 );
                 $schema = $this->_options['schema'];

            } else { // kind of assuming Mysql here, but the whole model does that.
                if ($this->_options['schema'] != $config->database->dbname) {
                    $initialize[] = sprintf(
                        $templateThis, 'setSchema', '"' . $this->_options['schema'] . '"'
                     );
                }
                $schema = $this->_options['schema'];
            }
        } else {
            $schema = $config->database->dbname;
        }
```
this `isset($this->_options['schema'])` still will be return `FALSE`

rianorie:
Why would it return false? You can define the schema on the commandline: `vendor/bin/phalcon.php model foobar --schema your_schema_goes_here`

vitaliykoziy:
If it will be return true, that is mean, in your old code never will be run this 

```
elseif ($adapter == 'Postgresql') {
            $schema = 'public';
            $initialize[] = sprintf(
                $templateThis, 'setSchema', '"' . $schema . '"'
            );
        } else {
            $schema = $config->database->dbname;
        }
```

and in new code never will be run this 
```
else {
            $schema = $config->database->dbname;
        }
```

I created this pull request, because for me this code `isset($this->_options['schema'])` returned `FALSE`

rianorie:
hm, alright. Well, it's merged.

I'm going to be looking at rewriting a bunch of this and I'll try to be
more mindful and be database agnostic rather than how we went about it the
first time, focusing primarily on MySQL.

Thanks for elaborating on the schemas :)

vitaliykoziy:
it is can be interesting.
http://forum.phalconphp.com/discussion/6034/connect-to-not-postgresql-public-schema
for You and for me.
Have a nice day :)

vitaliykoziy:
Please look heare
http://forum.phalconphp.com/discussion/3607/database-adapter-not-setting-schema-in-postgresql

if name of schema will be not ```public``` than 
```isset($this->_options['schema']) ``` will be return ```TRUE```, because for connect to another schema , ```schema ``` should identified like this ``` 'schema'   => 'app'``` (for example)

but if ``` 'schema'``` not identified in config file of database connection, for PostgreSql it should be ```public``` as default.

PR #270, opened 2014-12-03
Fix broken paths when using config.ini

stgatev:
When using config.ini (option --use-config-ini) the paths are set to one extra level up.


phalcon:
Thanks

PR #287, opened 2015-01-24
Fixes issue where model classes are not generated...

emilh91:
... even when "models" argument is still present.

The following error is thrown when the config file does not have an [application] section (and a "modelsDir" key in it):
Builder doesn't knows where is the models directory


phalcon:
Thanks

PR #298, opened 2015-01-30
Fixed model config path loading 

### LINKED ISSUE #278: create-model command fails to load config if directory option is not specified
I was trying to generate a model using the `phalcon model` command but despite having a valid database configuration that works with other phalcon commands (e.g. `phalcon migration generate`) I kept getting

> `Error: SQLSTATE[28000] [1045] Access denied for user 'root'@'localhost' (using password: YES)`

After several unsuccesfull attempts I decided to debug the issue.
First, the config that was loaded was not my user defined, but was instead one of the stubs from `vendor/phalcon/devtools/templates/project/simple/config.ini`. A quick look at **scripts/Phalcon/Builder/Component.php::_getConfig($path)** revealed that after not finding a config file under _app/config/_ or _config/_ the current directory is recursively searched for a config file and the one from the devtools templates inside my _vendor/_ directory is loaded. But why my user defined configuration couldn't be found in the first place? The reason was the way that **$path** parameter was concatenated with **$configPath**.

If 'directory' option is not set to the `phalcon model` command then we use the current directory - see **scripts/Phalcon/Commands/Builtin/Model.php::Line 242**:

``` php
} else {
    $path = '.';
}

$config = $this->_getConfig($path);
```

leading to a concatenated path `.app/config//config.php` which is definitely not a valid file since we are missing a '/' directory separator between the dot and the _app_ directory. A simple change to

``` php
    $path = './';
```

fixed the issue for me, but haven't tested for side effects.

7thcubic:


rianorie:
You would make me a very happy project collaborator if you could have sensible commit messages. Take a look at this: https://github.com/erlang/otp/wiki/Writing-good-commit-messages. I notice this repo doesn't have any set rules on this as of yet, but I'll get to work on that asap.

7thcubic:
fixed it all

sergeyklay:
@7thcubic  Can u please use here something like `realpath` and `DIRECTORY_SEPARATOR` ?

7thcubic:
#354

Issue #208, opened 2014-04-22
Webtools "Migrations" doesn't work on multi-module project


rsignavong:
Hi,

I've created a multi-module projet with devtools and enabled-webtools.

Then I changed application baseUri with '/' in /apps/frontend/config/config.php file.

When going to /webtools.php every link menu work like a charm except for Migrations "/webtools.php?_url=/migrations" that throw a php warning : 

Warning: mkdir(): No such file or directory in phalcon-devtools/scripts/Phalcon/Web/Tools/controllers/MigrationsController.php on line 34

The _getMigrationsDir() function seems not working for multi-modules project : 

 protected function _getMigrationsDir()
    {
        $migrationsDir = 'app/migrations';
        if (!file_exists($migrationsDir)) {
            mkdir($migrationsDir);
        }
        return $migrationsDir;
    }

I use Phalcon Devtools 1.3.1

PR #302, opened 2015-01-30
Added path's for module migrations 

7thcubic:
as well as fallback to project root directory. Issue #208


sergeyklay:
related #266

PR #303, opened 2015-01-31
Changed double-quotes into single quotes to avoid interpreting escape se...

pondermatic:
...quences.

Improved grammar.


rianorie:
awesome, merged! :)

PR #304, opened 2015-02-03
Fixed bug in db service closure.

pondermatic:
dbname was not using the correct config property.


rianorie:
I ran into a PR with the same fix for the `Module.php` before this one, so thanks, but it's already in there at this point :)

PR #306, opened 2015-02-04
Fix gen-stubs

brainformatik:
1) 
Boolean fix for broken acl class

```php
abstract class Acl {
    const ALLOW = 1;
    const DENY = ;
}
```

now its 
```php
abstract class Acl {
    const ALLOW = 1;
    const DENY = 0;
}
```

2)
Clean up indentation


phalcon:
We're now using:

```php
zephir stubs
```

PR #312, opened 2015-02-25
Fixes config path error when using INI files

Widcket:
Solves the "Database configuration cannot be loaded from your config
file" error in projects created with –-use-config-ini


sergeyklay:
@Widcket Fixed

PR #313, opened 2015-02-26
Automatic relationship detection

CodingBeard:
Detect hasMany and belongsTo relationships automatically and add them to the initialize function.


rianorie:
Thanks!

PR #315, opened 2015-03-01
Fix default path for Controllers creation

austerus:
This is a simple fix for an issue that prevents the ControllersController from loading the app config and using the configured controllersDir.

The default path is an empty string, therefore unusable to get the path to the config.
Since the root dir is "public", the simplest way is to go up one dir and append '/', allowing a proper load for the config file.


sergeyklay:
@austerus The PR needs to be rebased

sergeyklay:
Fixed in 2.0.x

PR #316, opened 2015-03-01
Add feature to reverse a migration

frohlfing:
As until now the up function is called if you migrate to a higher version. For the reverse migration the down function is called instead. In addition the morph function will be called in both cases (if existing).

### DIFF
--- a/scripts/Phalcon/Migrations.php
+++ b/scripts/Phalcon/Migrations.php
@@ -24,8 +24,6 @@ use Phalcon\Script\Color;
 use Phalcon\Version\Item as VersionItem;
 use Phalcon\Mvc\Model\Migration as ModelMigration;
 
-
-
 class Migrations
 {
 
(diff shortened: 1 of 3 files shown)


sergeyklay:
@gently69 The PR needs to be rebased

sergeyklay:
@rohfra The PR needs to be rebased

Green-Cat:
:+1: for rollbacks

TrogWarZ:
Rollback is good!

@frohlfing , why you remove checks for files existence and versions? Maybe, just check return value from up() method (or throw exception) and rollback by this flag?

sergeyklay:
@frohlfing ping

frohlfing:
Pong :-)

I'm very busy at the moment. I will be make a rebase next month.

sergeyklay:
@frohlfing

PR #317, opened 2015-03-04
better condition in scripts/Phalcon/Builder/Model.php

rzZzv:
I was getting the error:

screen shot 2015-03-04 at 18 29 50

### Note: phalcon-models is just an alias that I did to generate all the models and its relationships

, but this just happened on these last commits, I tried to understand the code, and came to the conclusion that on line 248 the condition becomes false, indeed, because `$this->_options['modelsDir']` is actually set but the path doesn't really exist.
The right path is set on line 254:

        $modelsDir = $config->application->modelsDir;

Then the result is,

screen shot 2015-03-04 at 18 42 43

# it's worth to mention, I tried it on a Mac and Gnu/Linux Machine (Ubuntu distro to be precise)...
Windows guys,...sorry.. I don't care...
---
Hope, it helps someone..I was getting in panic...

### DIFF
--- a/scripts/Phalcon/Builder/Model.php
+++ b/scripts/Phalcon/Builder/Model.php
@@ -245,10 +245,10 @@ class Model extends Component
 
         $config = $this->_getConfig($path);
 
-        if (!isset($this->_options['modelsDir'])) {
+        if (!isset($this->_options['modelsDir']) || !file_exists($this->_options['modelsDir'])) {
             if (!isset($config->application->modelsDir)) {
                 throw new BuilderException(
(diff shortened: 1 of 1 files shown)


rzZzv:
wow, it's worth mention btw, I just tried it on a Mac...

rzZzv:
Ok, I tried it on a Gnu/Linux machine (Ubuntu distro, to be precise) and the problem was the same, only after this little change I was able to generate the models and their relations.. Windows guys, I'm sorry..I don't care

sergeyklay:
Thanks

Zheness:
Thanks !

PR #324, opened 2015-03-15
improve scaffold with textarea on TEXT columns. Remove date type for TEXT

james2doyle:
For some reason the TEXT column was being assigned to a `type="date"` in the [Scaffold component](https://github.com/phalcon/phalcon-devtools/blob/221045950e70ed9674e3f6de9e10c630c858a693/scripts/Phalcon/Builder/Scaffold.php#L402). This change turns any TEXT column into a textarea field.

Results: 

```php
<?php echo $this->tag->textArea(array("column_name", "cols" => 30, "rows" => 4)) ?>
```


sergeyklay:
Thanks

PR #325, opened 2015-03-15
add support for enum in scaffolding

james2doyle:
Out of all the `Column::TYPE_*` methods, there is no `ENUM`. Doing some debugging, it looks like enum is represented as **5**. This PR adds support for adding a `staticSelect` field when an enum field is detected.

Here are the results for the PHP template. The field is served an empty array, not sure if there is a way to grab the field enum values, but that would be awesome.

```php
<?php echo $this->tag->selectStatic(array("column_name", array())) ?>
```


phalcon:
Thanks

PR #327, opened 2015-04-01
Add mapcolumn to all-models

cspeak:
Add the parameter --mapcolumn to all-models/create-all-models


sergeyklay:
@phalcon ping :)

Issue #339, opened 2015-04-27
Model Commad: Incorrect generation of relations if table name is used in other database.


escribiendocodigo:
Example:
```php
class Group extends \Phalcon\Mvc\Model {
    .
    .
    .
    public function initialize()
    {
        //Incorrect generation of relations
        $this->hasMany('id', 'Product', 'id_group', array('alias' => 'Product'));//table product in databaseA
        $this->hasMany('id', 'Product', 'id_group', array('alias' => 'Product'));//table product in  databaseB
        $this->hasMany('id', 'Provider', 'id_group', array('alias' => 'Provider'));//table provider in databaseC 
    }
    .
    .
    .
}
```

PR #340, opened 2015-04-27
Issue #339

escribiendocodigo:
Model Commad: Resolved generation of relations if table name is used in other database.


sergeyklay:
:+1:

Issue #342, opened 2015-04-28
NFR:  option to generate Abstract Class


escribiendocodigo:
in commands model and all-models add abstract option to generate Abstract Class

Example:
```
$ phalcon model product --abstract
```
```
$ phalcon all-models --abstract
```

Result:
```php
<?php

abstract class AbstractProduct extends \Phalcon\Mvc\Model
{
    .
    .
    .
}
```

PR #343, opened 2015-04-29
NFR Issue #342

escribiendocodigo:
In commands model and all-models added abstract option to generate Abstract Class


sergeyklay:
Thanks

PR #344, opened 2015-04-29
fix intellisense on magic methods

stefanchiriac:
In version 2.0 intellisense for magic methods stop working. That's because of the missing variables in devtools. This fixes the problem

http://forum.phalconphp.com/discussion/6520/phalcon-2-in-phpstorm


sergeyklay:
Thanks

PR #349, opened 2015-04-30
Properly spell DIRECTORY_SEPARATOR

josegonzalez:
This fixes issues where the constant is transformed into a string and used in concatenation :(


sergeyklay:
Thanks

PR #353, opened 2015-05-02
Better module routes, cleanups

7thcubic:
Allows modules to have different namespace, retrieves from className instead of standard namespace created by dev tools.


sergeyklay:
Thanks

PR #398, opened 2015-05-13
[2.0.x] Remove unused stubs

ogarbe:
in the 2.0.x branch, why preserving previous stubs ?


andresgutierrez:
When merging this to master it will remove those stubs

ogarbe:
but if you use devtools via composer you don't take master version you use a branch..

something like 

```json
"require-dev":{
        "phalcon/devtools":"1.3.x"
    }
```

or 

```json
"require-dev":{
        "phalcon/devtools":"2.0.x"
    }
```

logically nobody should require the master branch

andresgutierrez:
This is not mentioned in the docs or the README, everyone is expecting to find released versions in master as this is how it had worked since now

sergeyklay:
@ogarbe What benefits you will receive by removing all these stubs? Maybe your issue can be solved differently?

ogarbe:
in the readme the require is set to take the v2.0.1 :-)

bitwise, this change was accepted for 1.3.x branch and it wasn't a problem since https://github.com/phalcon/phalcon-devtools/tree/1.3.x

What's the advantage to have 10 versions of stubs in the same branch ? Stubs are used to help developers, if you use one version, you prefer the system to have the stubs for the version you use.

andresgutierrez:
@sergeyklay I think that should not be changed, we must use `master` for development and versioned branches for development

ogarbe:
@sergeyklay the benefit is quite clear, when loading devtools in my env for one version, i only have one version of the stub file instead of having more than 10. It's quicker for my development environment and it's easier for me as i don't risk to use a bad one.

@andresgutierrez i suggest to use versioned branch as it's used in packagist / composer : one branch for a major - minor version. Using this you can have evolution on two major version (2.0 and 2.1 for example), and with this people can follow. If you do this, having stubs for 2.0 in the 2.0 version is logical, and stubs for 2.1 in the 2.1 also. At the end, it helps people.

andresgutierrez:
@ogarbe packagist/composer can work with both branches and tagged versions, I don't see why someone needs to change to one version or another, if he has applications running on multiple versions does he need to clone the repo many times?

sergeyklay:
@andresgutierrez Fixed

andresgutierrez:
@sergeyklay Thanks :+1:

ogarbe:
in my case, i'm working on changing an evolving app from 1.3.4 to 2.0.x, it's not easy and can take some time. At the same time the application evolves hence a lot of rebases.

So having a devtool pointing to 1.3.x version will give me all stubs for the 1.3.x code, and my env is setup automatically. Then i switch to phalcon 2.0, change the require for devtool to 2.0.x and the stubs follows.

Then in 3 month you'll publish 2.1 version of phalcon with new stubs, when i'll migrate, i'll use devtool in 2.1.x branch hence stubs will follow.

Concerning the master branch, as it follows phalcon branch also, it seems quite logical stubs are for the last publish version of phalcon (hence the 2.0.1 at this time).

virgofx:
Any chance we can get this merged ?

sergeyklay:
@andresgutierrez ?

virgofx:
Please merge :+1:

ogarbe:
:up: any one ?

sergeyklay:
We're here :) I can't decide, becaus there are no enough quorum

rianorie:
@sergeyklay is there a vote taking place for this?

sergeyklay:
@rianorie It is here

andresgutierrez:
I've used PHPStorm and Netbeans and you don't have to add the whole directory that includes every version stubs but just the directory having the version used in that project. Not sure if that is related to the IDE/Editor you are using.

rianorie:
my two cents: get rid of anything before 1.3.4, leaving that branch in place to facilitate people using that version of the framework. Move the master along with the master of the framework and tag the master with the version as name every time it's updated to reflect the new framework version.

I can tell PhpStorm to exclude/ignore the old stub folders, I can also composer require a specific version, but to me it seems to make more sense to do it like this, because whatever way you use the dev tools like this you'll always just have one version of the stubs in your project.

ogarbe:
@andresgutierrez yes you can exclude some directories from the ide, but what's the use ? if devtools is tagged/branched with version like cphalcon is, why would i (or anybody else) use stubs from another version ?

ogarbe:
another advantage would be that we could have only one dir for stubs, which is easier to see diff between one version and another, which can also help to do upgrades.

f3l1x:
:+1:

ogarbe:
<a href="http://www.youtube.com/watch?feature=player_embedded&v=aHN6AViJAvI
" target="_blank">IMAGE ALT TEXT HERE</a>

sergeyklay:
We're here

virgofx:
@andresgutierrez Including the specific version is not acceptable as it would require updating project files every time a composer update occurs. The better solution would be to just specify the version in the composer.json and then have one's IDE pick it up automatically. This is how all other stubs work. Why is this still up for discussion? We shouldn't cater to bad practice just because it was how it was done originally.

Please merge such that the stubs only contain the relevant version in sync with cphalcon version.

cnelissen:
+1 for merging this. If you install Phalcon 2.0.8 and then require phalcon/devtools 2.0.8, there is no reason to have stubs for any other version. Devtools master should just track phalcon master, whatever API is there is what is current, older versions exist at tagged releases... There is no reason to have 25 versions of stubs when you can only have a single version of Phalcon installed at a time...

ogarbe:
hello, can we have a decision on this ?

sergeyklay:
Sure

virgofx:
+1

lbaroVPG:
+1

KASTINpl:
+1

sergeyklay:
@ogarbe https://forum.phalconphp.com/discussion/10903/what-do-you-think-we-should-do-with-the-phalcon-stubs

ogarbe:
Thanks @sergeyklay but the poll isn't working ;-) !

sergeyklay:
@ogarbe Working. Don't worry. Some issue with cache which be fixed soon.

lbaroVPG:
Thank you @sergeyklay!

virgofx:
@sergeyklay The community has spoken via the poll.  Is it possible to move all stubs into single folder, one copy, and start tagging releases corresponding to Phalcon now?

sergeyklay:
Yes. Could you please do PR?

virgofx:
Are you okay with leaving existing tags in place such that separated versions will be applicable from this point forward on both 2.0.x and 2.1.x branches? Would be way complicated and difficult to go backwards in time and adjust for previous tags/revisions.  At @ogarbe Will need to do for both current branches (2.0.x and 2.1.x)

ogarbe:
it's done, for 2.0.x and 2.1.x 

alls stubs are now in stubs dir

sergeyklay:
Thanks

stefanchiriac:
this decision is great for development, but really awful for maintaining applications on different versions. What will happen in the future with versions like 2.1.1, 2.1.2 and so on? Everything will be on 2.0.x? Doesn't make sense at all since 2.1.2 might have some extra functionality compared to 2.1.1.
I really hope I'll still have the older minor versions and not being forced on the last one. I wouldn't say that if Phalcon wouldn't had a great history of new features and incompatibilities between minor versions.

ogarbe:
@stefanchiriac : if you want to have stub for 2.1.1 you just set your composer to use v2.1.1 instead of 2.1.x

this change is exactly for permitting to work on different versions at the same time ;-)

stefanchiriac:
then we should start making branches for every minor version. But this might be an overkill later on. Doesn't sound such a great idea. A middle way would be that each major version to have the stubs for all minor versions.

sergeyklay:
> then we should start making branches

tags. not branches
https://github.com/phalcon/phalcon-devtools/releases

PR #399, opened 2015-05-13
Fixed Composer instruction

sergeyklay:


sergeyklay:
Also for 2.0.x to avoid confusion

PR #401, opened 2015-05-14
fix DbAdapter param according to expected value

dadagama:
**DbAdapter** expected as param something like this
```
    array(
        'adapter'    => 'Mysql',
        'host'       => 'localhost',
        'username'   => 'root',
        'password'   => 'root',
        'dbname'     => 'my_db_name',
        'charset'    => 'utf8',
    );
```
without **database->** part, it is passing the entire config array and when we try to execute a phql statement/query, it throws a php notice like this:
```
Notice: Array to string conversion in config/services.php on line 32
```


sergeyklay:
Thanks

PR #409, opened 2015-05-18
bootstrap has no alert-error class + add warning class

iGusev:


sergeyklay:
Can you please send PR to 2.0.x?

iGusev:
Oh. Its [already fixed](https://github.com/phalcon/phalcon-devtools/blob/2.0.x/scripts/Phalcon/Web/Tools.php#L312) in 2.0.x

PR #410, opened 2015-05-18
Added support for micro-app config directory

cspeak:
A micro-application's config file is located in "config/config.[ini|php]. The current devtools only supports "app/config/config.[ini|php]". I added support for the micro-app config directory.

### DIFF
--- a/scripts/Phalcon/Commands/Builtin/AllModels.php
+++ b/scripts/Phalcon/Commands/Builtin/AllModels.php
@@ -68,12 +68,25 @@ class AllModels extends Command
 
         $config = null;
         if (!$this->isReceivedOption('models')) {
-            $fileType = file_exists($path . "app/config/config.ini") ? "ini" : "php";
+          
+            // Check in the app config directory but also check in the micro-app config directory
+            if(file_exists($path . "app/config/config.ini") || file_exists($path . "config/config.ini")){
+                $fileType = 'ini';
+            } else {
+                $fileType = 'php';
+            }
 
             if ($this->isReceivedOption('config')) {
                 $configPath = $path.$this->getOption('config')."/config.".$fileType;
             } else {
-                $configPath = $path."app/config/config." . $fileType;
+                // Check for the app config files
+                if(file_exists($path."app/config/config.".$fileType)){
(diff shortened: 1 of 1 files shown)


sergeyklay:
@cspeak Can you please create PR to 2.0.x branch?

scripts/Phalcon/Commands/Builtin/AllModels.php
```diff
@@ -68,12 +68,25 @@ public function run($parameters)
 
         $config = null;
         if (!$this->isReceivedOption('models')) {
-            $fileType = file_exists($path . "app/config/config.ini") ? "ini" : "php";
+          
+            // Check in the app config directory but also check in the micro-app config directory
+            if(file_exists($path . "app/config/config.ini") || file_exists($path . "config/config.ini")){
```
quantumpacket: What was the reason for not using the ternary as like before?

```php
$fileType = (file_exists($path . 'app/config/config.ini') || file_exists($path . 'config/config.ini'))
    ? 'ini'
    : 'php';
```

cspeak: Honestly, that is how I had it initially and just opted to display the logical flow. I'll close the pull requests and update to shorthand.

scripts/Phalcon/Commands/Builtin/AllModels.php
```diff
@@ -68,12 +68,25 @@ public function run($parameters)
...
+                $fileType = 'ini';
+            } else {
+                $fileType = 'php';
+            }
 
             if ($this->isReceivedOption('config')) {
                 $configPath = $path.$this->getOption('config')."/config.".$fileType;
             } else {
-                $configPath = $path."app/config/config." . $fileType;
+                // Check for the app config files
```
quantumpacket: Hopefully I got the logic down correct, a bit lacking in sleep :P
I would write it this way to keep it DRY. It should have the same functionality:

```php
if (file_exists($path . 'app/config/config.' . $fileType)) {
    $path = $path . 'app/'; // default is micro-app as shown below, but we add single/multi module dir here
}

$configPath = $path . 'config/config.' . $fileType;
```

PR #411, opened 2015-05-18
Cspeak patch - fix related model names to camelize correctly

cspeak:


cspeak:
This pull request was only supposed to contain the referenced model name camelize issue. I opened a different pull request initially to fix the micro-app config directory/file issue. But it looks like they're both present in here

sergeyklay:
@cspeak Can you please create PR to 2.0.x branch?

PR #412, opened 2015-05-19
FIxed camelize issue with alias and reference table names

cspeak:
When generating models, related Model names are not always camelized. This fixes that issue.


sergeyklay:
Thanks

PR #413, opened 2015-05-19
Error fix

uonick:
`SQLSTATE[HY000] [1045] Access denied for user ''@'localhost' (using password: NO)`
Error fix


sergeyklay:
@uonick Thanks for contributing!

`master` is freezed. Create PR to 2.0.x

PR #416, opened 2015-05-31
ide stubs for v2.0.2

denji:


sergeyklay:
@denji Thanks for contributing!

`master` is freezed. Create PR to `2.0.x`

Issue #425, opened 2015-06-14
Installing via composer on local machine may fail


ryomo:
To use IDE stubs, I tried to install devtools locally.
```sh
composer global require phalcon/devtools
```
But failed because devtools depends on ext-phalcon which is not needed on local machine.
(Of course ext-phalcon is needed on the server or virtual machines.)

---
Below should work, but this may ignore other dependencies. :worried: 
```sh
composer global require phalcon/devtools --ignore-platform-reqs
```

PR #426, opened 2015-06-14
#425 Remove requirement ext-phalcon

ryomo:


sergeyklay:
Why?

sergeyklay:

ryomo:
>Why?
Because someone like me install ext-phalcon not on the local machine but on virtual machines.
IDE is installed on the local machine.
(issue #425 )
Sorry for my bad English.

>We don't add changes directly to master, could you please submit this to the branch 2.0.x?
I made new PR( #428 ).

virgofx:
The current master still has ext-phalcon as a requirement. Devtools should not require this as it does prohibit installations for composer where IDEs and development environments which to include this as a development dependency.

sergeyklay:
use `php composer.phar install --ignore-platform-reqs`

PR #428, opened 2015-06-14
#425 Remove requirement ext-phalcon

ryomo:
To fix issue #425 .


sergeyklay:
Ho about?

```sh
php composer.phar update --ignore-platform-reqs
```

PR #432, opened 2015-06-19
Fix --excludefields option

ryo-utsunomiya:
Currently, --excludefields works only with --get-set option.
This PR fixes --excludefields behavior and enables to exclude fields without --get-set option.


sergeyklay:

ryo-utsunomiya:
I created https://github.com/phalcon/phalcon-devtools/pull/433. (I don't know how to change base branch...)
This PR will close.

PR #433, opened 2015-06-20
Fix phalcon model --excludefields option

ryo-utsunomiya:
Currently, --excludefields works only with --get-set option.
This PR fixes --excludefields behavior and enables to exclude fields without --get-set option.
This is re-submit of https://github.com/phalcon/phalcon-devtools/pull/432.


sergeyklay:
Thanks

PR #439, opened 2015-07-05
Adapt the php-dev-tools to work with oldest version of PHP < 5.4

RaphaelNeumann:
A small change that allows operation even in older versions, instantiating the class before calling the method of the same.

### DIFF
--- a/scripts/Phalcon/Web/Tools.php
+++ b/scripts/Phalcon/Web/Tools.php
@@ -413,9 +413,16 @@ class Tools
             throw new \Exception('Document root cannot be located');
         }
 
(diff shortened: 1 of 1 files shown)


sergeyklay:

sergeyklay:
@RaphaelNeumann

RaphaelNeumann:
Ok! I do this (https://github.com/phalcon/phalcon-devtools/pull/443) and now i see other lines need fix i will send other pull request soon :) hahahaha

sergeyklay:
I close this in accordance with #443

PR #443, opened 2015-07-10
Adapt the php-dev-tools to work with oldest version of PHP < 5.4

RaphaelNeumann:
A small change that allows operation even in older versions, instantiating the class before calling the method of the same.

**Resend on branch


sergeyklay:
@RaphaelNeumann can I merge this?

RaphaelNeumann:
of course!

sergeyklay:
Thanks

PR #449, opened 2015-07-16
Profiler documentation for getProfiles method fix

uLow:
In fact that getProfiles returns all the processed profiles as an array of \Phalcon\Db\Profiler\Item


sergeyklay:
Create PR/Issue in Phalcon/Zephir repo.
Devtools repo does only contains stubs, not edit. Especially in the `master` branch

uLow:
Then there is something wrong with stub generation I guess. As you can see, there are Item[] in https://github.com/phalcon/cphalcon/blob/master/phalcon/db/profiler.zep#L156

PR #450, opened 2015-07-16
getServices() documentation fix

uLow:
getServices() returns array of \Phalcon\Di\ServiceInterface


andresgutierrez:
IDE PHP stubs files are auto-generated from the Zephir sources here: https://github.com/phalcon/cphalcon/tree/2.0.x/phalcon

Submit changes directly in the IDE PHP stubs will be likely lost when the next regeneration is made

PR #451, opened 2015-07-24
Update Manager.php

lazycommit:
just an arg


sergeyklay:
@lazycommit Can you pleace create PR for 2.x ? (for all Phalcon's stub version where this need)

lazycommit:
We do not, actually, use 2.x. But I'll look up if time free happens;)
24 июля 2015 г. 17:09 пользователь "Serghei Iakovlev" <
notifications@github.com> написал:

PR #452, opened 2015-07-24
Query setBindParams add merge param

alexivenkov:


andresgutierrez:
It's already there: https://github.com/phalcon/phalcon-devtools/blob/2.0.x/ide/2.0.5/Phalcon/mvc/model/Query.php#L403

PR #458, opened 2015-07-30
fixed composer error

qetee:
cannot find the box from packagist.org; the correct is "kherge/box", but not just "box".


denji:
Yes the vendor prefix is required to search the registry composer :+1:

sergeyklay:
We don't add changes directly to `master`. Create PR to `2.0.x` please

sergeyklay:
@qetee

PR #460, opened 2015-07-30
fixed the path error of webtools page

qetee:
the correct path is the path of file "webtools.php",  not the path of the project root.   when i create project with devtools and enable webtools,  the "webtools.php" file is in the public dir.


sergeyklay:
We don't add changes directly to `master`. Create PR to `2.0.x` please

sergeyklay:
@qetee

PR #464, opened 2015-08-02
[ PHPDoc ] Phalcon\Di::get return mixed value

ph55:
No returning value for Di class in PHPDoc


sergeyklay:
We don't add changes directly to `master`.

IDE PHP stubs files are auto-generated from the Zephir sources here: https://github.com/phalcon/cphalcon/tree/2.0.x/phalcon

Submit changes directly in the IDE PHP stubs will be likely lost when the next regeneration is made

Issue #465, opened 2015-08-03
Mysql Migration BIGINT


Surt:
It seems as defined in Phalcon as Column::BIND_PARAM_INT
but when trying to generate a migration (using phalcon-devtools master branch) it says : 

``` Error: Unrecognized data type 14 at column .... ```

The type 14 is BIGINT on Phalcon but definition["bindType"] is INT and is not getting it.

```php
/**
 * Smallint/Bigint/Integers/Int are int
 */
if memstr(columnType, "bigint") {
    let definition["type"] = Column::TYPE_BIGINTEGER,
        definition["isNumeric"] = true,
        definition["bindType"] = Column::BIND_PARAM_INT;
    break;
}
```

Any workaround or fix?

by the way, wouldn't be easier to use an array to map Types and characteristics instead of those constants so difficult to mantain?

stormweaver:
Stumbled onto the same issue today.
Fixed easily by adding

```php
case Column::TYPE_BIGINTEGER:
			        $fieldDefinition[] = "'type' => Column::TYPE_BIGINTEGER";
			        break;
```
here:  
https://github.com/phalcon/phalcon-devtools/blob/master/scripts/Phalcon/Mvc/Model/Migration.php#L235
just before the default switch.

sergeyklay:
Can you please create PR?

PR #466, opened 2015-08-04
Added support for bigint field type during "phalcon migration generate".

stormweaver:
Following up on this issue: https://github.com/phalcon/phalcon-devtools/issues/465


sergeyklay:
We don't add changes directly to `master`. Can you please create PR to `2.0.x`?

sergeyklay:
Close in accordance with #467

PR #467, opened 2015-08-05
Added support for bigint field type during "phalcon migration generate".

stormweaver:
Following up on bug #465 and commit #466 but targeting 2.0.x branch.
Sorry 'bout the wrong branch, still new here. :)


sergeyklay:
Thanks

PR #498, opened 2015-09-15
Send router back to DI

bmoore:
The routes set here are not actually being set, it just coincidentally works with a single demo module due to the "defaultNamespace" set in services.php


sergeyklay:
@bmoore We don't add changes directly to master. Can you please create PR to 2.0.x?

Issue #524, opened 2015-10-11
PostgreSQL migrations doesn't use schema name


Zaszczyk:
I run phalcon migration run with such migration file:
```
<?php 

use Phalcon\Db\Column;
use Phalcon\Db\Index;
use Phalcon\Db\Reference;
use Phalcon\Mvc\Model\Migration;

/**
 * Class ErrorsMigration_101
 */
class ErrorsMigration_101 extends Migration
{
    /**
     * Define the table structure
     *
     * @return void
     */
    public function morph()
    {
        $this->morphTable('errors', array(
                'columns' => array(
                    new Column(
                        'id',
                        array(
                            'type' => Column::TYPE_INTEGER,
                            'default' => "nextval('errors_id_seq'::regclass)",
                            'notNull' => true,
                            'autoIncrement' => true,
                            'first' => true
                        )
                    ),
                    new Column(
                        'name',
                        array(
                            'type' => Column::TYPE_VARCHAR,
                            'notNull' => true,
                            'size' => 32,
                            'after' => 'id'
                        )
                    ),
                    new Column(
                        'type',
                        array(
                            'type' => Column::TYPE_INTEGER,
                            'notNull' => true,
                            'after' => 'name'
                        )
                    ),
                    new Column(
                        'content',
                        array(
                            'type' => Column::TYPE_TEXT,
                            'notNull' => true,
                            'size' => 1,
                            'after' => 'type'
                        )
                    ),
                    new Column(
                        'created_at',
                        array(
                            'type' => Column::TYPE_INTEGER,
                            'notNull' => true,
                            'after' => 'content'
                        )
                    )
                ),
                'indexes' => array(
                    new Index('errors_pkey', array('id'))
                ),
            )
        );
    }

    /**
     * Run the migrations
     *
     * @return void
     */
    public function up()
    {

    }

    /**
     * Reverse the migrations
     *
     * @return void
     */
    public function down()
    {

    }

}
```

and I got errors:
```
1444565584.0848: SELECT CASE WHEN COUNT(*) > 0 THEN 1 ELSE 0 END FROM information_schema.tables WHERE table_schema = 'dbName' AND table_name='errors'  => 1444565584.0933 (0.0085380077362061)
[...]

PR #526, opened 2015-10-11
fix table path for PostgreSQL

Zaszczyk:


scripts/Phalcon/Mvc/Model/Migration.php
```diff
@@ -493,7 +493,12 @@ public static function migrate($fromVersion, $toVersion, $tableName)
...
     public function morphTable($tableName, $definition)
     {
-        $defaultSchema = self::$_databaseConfig->get('dbname');
+        if (self::$_databaseConfig->get('adapter') === 'Postgresql') {
+            $defaultSchema = self::$_databaseConfig->get('dbname') . '.' . self::$_databaseConfig->get('schema');
+        } else {
+            $defaultSchema = self::$_databaseConfig->get('dbname');
+        }
+
         $tableExists = self::$_connection->tableExists($tableName, $defaultSchema);
```
sergeyklay: See https://github.com/phalcon/cphalcon/blob/master/phalcon/db/dialect/postgresql.zep#L535

Zaszczyk:
Did You see issue: https://github.com/phalcon/phalcon-devtools/issues/524? This pull request is connected with that.

sergeyklay:
But this solution is not flexible enough (I mean MySQL, SQLite, etc.). 
Also please see this Postgres-like case:
```php
$config = [
    'database' => [
        'adapter'  => 'Postgresql',
        'host'     => 'localhost',
        'username' => 'username',
        'password' => '********',
        'dbname'   => 'dbname',
        'port'     => 5432,
    ],
    // ...
];
```
Without schema

sergeyklay:
Fixed in #528

PR #541, opened 2015-10-21
Add standard error output

ryomo:
### Currently
`phalcon command` outputs STDOUT(standard output) even if it throws the exception.

So below doesn't work correctly.
```sh
$ phalcon invalid-command 2> stderr.log
```

### Should
outputs STDERR(standard error) when an error happened.


sergeyklay:
Thanks

PR #572, opened 2015-12-06
trying to load  the abstract base model class file with —extends para…

zhao5908:
Repeating the generated will trigger fatal error: Class 'x' not found in x on line x,when use —extends=BaseModel


sergeyklay:
We don't add changes directly to `master` branch. Can you please do PR to `2.0.x` branch?

zhao5908:
Thanks for your help , I created a new pr.

PR #574, opened 2015-12-06
2.0.x trying to load the abstract base model class file with —extends 

zhao5908:
Repeating the generated will trigger fatal error: Class 'x' not found in x on line x,when use —extends=BaseModel


scripts/Phalcon/Builder/Model.php
```diff
@@ -336,6 +336,12 @@ public function build()
 
                 $possibleMethods['getSource'] = true;
 
+                $extends = $this->options->get('extends', '\Phalcon\Mvc\Model');
+                $extendsPath = dirname($modelPath) . '/'. $extends . '.php';
```
sergeyklay:
```php
// For example \Project\Model\Base
$extends = $this->options->get('extends', '\Phalcon\Mvc\Model');

// dirname($modelPath) + '/' + ' \Project\Model\Base' . '.php'
$extendsPath = dirname($modelPath) . '/'. $extends . '.php';
```

???

zhao5908: ohh, Let me think about it again。

PR #575, opened 2015-12-08
add user custom autoload

zhao5908:
In order to execute the following command can be repeated

phalcon model --namespace=Service\\ORM --extends=Service\\ORM\\BaseModel --get-set --force --name=tablename

user custom BaseModel class  needs to be loaded


sergeyklay:
:+1:

PR #578, opened 2015-12-17
Phalcon\Mvc\Model::find docblock bug

cnelissen:
Docblock is formatted incorrectly for the Phalcon\Mvc\Model::find method.


sergeyklay:
We don't add changes directly to `master` branch. Can you please do PR to `2.0.x` branch?

PR #591, opened 2016-01-16
using defined const for Column:TYPE_TIMESTAMP

mattvb91:


sergeyklay:
I left this here because older versions may not have this constant

mattvb91:
@sergeyklay do we need a 2.1.x branch for changes like this instead?

sergeyklay:
@mattvb91 Can you please send PR to the `2.1.x` branch?

mattvb91:
Thanks @sergeyklay have opened new PR

PR #609, opened 2016-02-18
Options to store information about last migration in database

rusAnthrax:
Much clearer way to store migrations log. Automatically solve problems with deployment to a number of servers which use same db servers.
Also some PSR1/PSR2 cosmetic changes.


sergeyklay:
Thanks

Green-Cat:
:+1:

PR #610, opened 2016-02-18
Rename migrations

zinovyev:
Two specific flags added:

- --ts-base (timestap based) for "migration run" action
- --descr (description) for "migration generate" action.

The migrations folder will be called named the timestamp based mode  _1455789042106821_new_name_, where "1455789042106821" is the current timestamp (in microseconds) and "new_name" is the value of the --descr option.
When you're working in team on a big project, it is a convinient way to name migrations after current timestamp value, thus your migrations wouldn't be overriden.


sergeyklay:
Could you please submit just the commits that belong to the PR?

sergeyklay:
I think 2.1.x branch is best  place  for this PR

zinovyev:
Ok. I'll fix it to be able to commit changes to the 2.1.x branch. Thanx for this suggestion.

zinovyev:
Moved to https://github.com/phalcon/phalcon-devtools/pull/611

PR #611, opened 2016-02-19
Rename migrations

zinovyev:
Two specific flags added:

    --ts-base (timestap based) for "migration run" action
    --descr (description) for "migration generate" action.

The migrations folder will be called named the timestamp based mode 1455789042106821_new_name, where "1455789042106821" is the current timestamp (in microseconds) and "new_name" is the value of the --descr option.
When you're working in team on a big project, it is a convinient way to name migrations after current timestamp value, thus your migrations wouldn't be overriden.

P.S. Moved from PR to 2.0.x by the suggestion of @sergeyklay


rusAnthrax:
:+1:

scripts/Phalcon/Migrations.php
```diff
@@ -173,63 +211,134 @@ public static function run(array $options)
...
+     * @param ItemInterface $version
+     * @param string        $startTime
+     */
+    public static function setCurrentVersion(array $options, ItemInterface $version, $startTime = 'NOW()')
+    {
+        if (isset($options['migrationsInDb']) && (bool)$options['migrationsInDb']) {
+            /** @var AdapterInterface $connection */
+            $connection = self::$_storage;
+            // TODO: TRUNCATE to be removed on refactor
+            $connection->execute('TRUNCATE TABLE `phalcon_migrations`;');
```
sergeyklay: ?

rusAnthrax: For now we keep only one (the last) migration in this table. It is a default behavior, see how it is done in "else" section with file storage.
In a mater of days there will be a refactored version which stores full log of completed migrations, so this line will be gone.

sergeyklay:
@zinovyev Could you please rebase onto `2.1.x`?

zinovyev:
It was pretty hard to rebase on this step. So I just merged the upstrem branch to my local.
Is that ok?

sergeyklay:
Ok

sergeyklay:
Thanks

PR #615, opened 2016-02-23
fix #419 (branch 2.0.x)

### LINKED ISSUE #419: Phalcon migration Generate/Run error with Postgresql
Generate and run success with MySql but error on Postgres with SCHEMA.

I think team add new video demo for Postgres sample as Mysql (Youtube: https://www.youtube.com/watch?v=Vx-9GiZt68I)

Daredzik:
Phalcon migration Generate/Run error with Postgresql #419 fixed.

only crash on Postgresql


sergeyklay:
Thanks

PR #616, opened 2016-02-23
Prevents double firing methods down() on downgrade.

reczu:


sergeyklay:
We don't add changes directly to `master` branch. Can you please do PR to the `2.0.x` branch?

Issue #520, opened 2015-10-02
Migration rollback doesn't work


evan-mark4s:
Test case: I have a migration version 1.0.2 which contains 1 file: accounts.php, then I have migration version 1.0.3 which contains 1 file: admins.php with functions up() and down(), no morph().

I try migrating back from 1.0.3 to 1.0.2 using `phalcon migration run --version=1.0.2` but the down() function is not called. I tried inserting a small 
```
echo "$fromVersion -> $toVersion ($tableName)\n";
```
on function migrate() in Mvc\Model\Migrations.php and this is what I got
```
1.0.3 -> 1.0.3 (admins)
1.0.3 -> 1.0.2 (accounts)
```
So it seems the rollback for `admins` table was never called. Please check.

sergeyklay:
What kind of rollback you are mean? Could you please provide migrations to reproduce?

joel-depiltech:
Up !
I have the same issue. Do you have a solution for me ?

sergeyklay:
Can you please provide script to reproduce?

joel-depiltech:
This is my migration file : 
<?php 

use Phalcon\Db\Column;
use Phalcon\Db\Index;
use Phalcon\Db\Reference;
use Phalcon\Mvc\Model\Migration;

class CiUsersMigration_101 extends Migration
{
    public function up()
    {
       self::$_connection->update(
            "ci_users",
            array("username"),
            array("Astro Boy"),
            "id = 2"
        );
    }

    public function down()
    {
        self::$_connection->update(
            "ci_users",
            array("username"),
            array("Paul"),
            "id = 2"
        );
    }
}

When I upgrade to version 1.0.1, the username of id=2 changed to "Astro Boy" , but when I downgrade to  version 1.0.0 the username never changed to "Paul"

joel-depiltech:
Up ! Please do you have an issue for me ?

sergeyklay:
I'll try to deal with this in the near future

sergeyklay:
@evan-mark4s @joel-depiltech Can you please check `2.0.x` branch?

joel-depiltech:
Thanks. I will check on next Tuesday. I will come back to you...

sergeyklay:
@joel-depiltech

sergeyklay:
I think issue is fixed `2.0.x` branch. This will be opened if the issue is actual

PR #617, opened 2016-02-23
Prevents double firing methods down() on downgrade.

reczu:


sergeyklay:
This constant introduced to solve #520 issue
See: bdd0266f19d42221882124970c90f179ad40fab6

reczu:
I have migration 1.0.1 with file test.php with up and down methods no morph

```php
...

   /**
     * Run the migrations
     *
     * @return void
     */
    public function up()
    {
        self::$_connection->addColumn('test', null, new Column('new_col', array('type' => Column::TYPE_VARCHAR, 'size' => 50,)));
    }

    /**
     * Reverse the migrations
     *
     * @return void
     */
    public function down()
    {
        self::$_connection->dropColumn('test', null, 'new_col');
    }
...
```

So if i try to rollback to version 1.0.0 i get following error:

```
Phalcon DevTools (2.0.10)

1456237484.2396: ALTER TABLE "test" DROP COLUMN "new_col"  => 1456237484.2424 (0.0027918815612793)
                                                    
  Success: Version 1.0.1 was successfully migrated  
                                                    
1456237484.243: ALTER TABLE "test" DROP COLUMN "new_col"ERROR: SQLSTATE[42703]: Undefined column: 7 ERROR:  column "new_col" of relation "test" does not exist
```

As you can see it fires method down twice and breaks migration.

Daredzik:
+1

sergeyklay:
I'll try to deal with this in the near future. Ping me if I forget

mokeev1995:
@sergeyklay ping

sergeyklay:
@sergeysviridenko Could you please take a look

sergeyklay:
Fixed in the `3.2.x branch`. Feel free to open a new issue if the problem appears again. Thank you for contributing.

PR #623, opened 2016-02-26
timestamped migrations fix and new config variables

rusAnthrax:
* introduced "logInDb" and "migrationsTsBased" config variables which may be overridden by cli options
* fixed very initial migration numbering in timestamped migration scheme


sergeyklay:
Thanks

PR #638, opened 2016-03-30
Remove semicolons from queries so it works with SQLite adapter.

basilfx:
Using PDO SQLite adapter and and the option `migrationsInDb = true` when applying migrations, I get query errors because the queries that create/select/insert `phalcon_migrations` end with a semicolon.

This PR removes the semicolons, which should not affect further operations. 

Details:

* Phalcon 2.1.x
* Phalcon Devtools 2.1.x
* Ubuntu 15.10
* PHP 5.6.11
* MySQL 5.6.28 

Logs:

This is what is printed to console.

With PDO SQLite, without fix:

```
PHP Fatal error:  Uncaught exception 'PDOException' with message 'SQLSTATE[HY000]: General error: 1 near ";": syntax error' in /opt/backend/vendor/phalcon/devtools/scripts/Phalcon/Migrations.php:436
```

With PDO MySQL, without fix:

```
1459337109.4718: SELECT IF(COUNT(*) > 0, 1, 0) FROM `INFORMATION_SCHEMA`.`TABLES` WHERE `TABLE_NAME`= 'projects' AND `TABLE_SCHEMA` = 'dfdb'  => 1459337109.4724 (0.00060606002807617)
```

With PDO SQLite, with fix:

```
1459337106.1921: SELECT CASE WHEN COUNT(*) > 0 THEN 1 ELSE 0 END FROM sqlite_master WHERE type='table' AND tbl_name='projects'  => 1459337106.1924 (0.00031399726867676)
```

### DIFF
--- a/scripts/Phalcon/Migrations.php
+++ b/scripts/Phalcon/Migrations.php
@@ -400,7 +400,7 @@ class Migrations
             }
 
             if (!self::$_storage->tableExists('phalcon_migrations')) {
-                self::$_storage->execute("CREATE TABLE `phalcon_migrations` (`version` VARCHAR(255), `start_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP, `end_time` TIMESTAMP NOT NULL DEFAULT '0000-00-00 00:00:00' NOT NULL);");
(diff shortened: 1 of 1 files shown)


basilfx:
Closed in favor of #639.

PR #639, opened 2016-03-30
Improve support for PDO Sqlite migrations

basilfx:
I know PDO SQLite migrations are not fully working (due to ALTER table limitations), but the following three commits should at least improve support for running migrations (I successfully converted structure from MySQL to SQLite).

* 2866803 fixes the issue described in PR #638.
* 3f543dc changes MySQL's `NOW()` in PHP's `date()`. The latter is already used in `Migrations.php`.
* 7e2c1cd fixes the issue where the file path is added as schema (see below).

Example of issue the last commit fixes (note the file name is used as schema in the CREATE TABLE statement):

```
1459346099.167: CREATE TABLE "/opt/backend/app/../data/df.db"."projects" (  `ID` INTEGER NOT NULL,  `name` VARCHAR(128) NOT NULL,  `userID` INTEGER NOT NULL,  `defaultQuestionID` INTEGER,  `access` VARCHAR(512),  `metadata` TEXT NOT NULL,  `created` DATETIME NOT NULL,  `modified` DATETIME,  `deleted` DATETIME,  PRIMARY KEY ("ID"),  CONSTRAINT `projects_ibfk_1` FOREIGN KEY ("userID") REFERENCES `users`("ID"),  CONSTRAINT `projects_ibfk_2` FOREIGN KEY ("defaultQuestionID") REFERENCES `questions`("ID") )PHP Fatal error:  Uncaught exception 'PDOException' with message 'SQLSTATE[HY000]: General error: 1 unknown database "/opt/backend/app/../data/df.db"' in /opt/backend/vendor/phalcon/devtools/scripts/Phalcon/Mvc/Model/Migration.php:663
```

I have also a patch ready that converts the missing column types for SQLite (Blob, Double, etc). I'll contribute this to CPhalcon.

### DIFF
--- a/scripts/Phalcon/Migrations.php
+++ b/scripts/Phalcon/Migrations.php
@@ -400,7 +400,7 @@ class Migrations
             }
 
             if (!self::$_storage->tableExists('phalcon_migrations')) {
-                self::$_storage->execute("CREATE TABLE `phalcon_migrations` (`version` VARCHAR(255), `start_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP, `end_time` TIMESTAMP NOT NULL DEFAULT '0000-00-00 00:00:00' NOT NULL);");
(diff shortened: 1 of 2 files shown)


sergeyklay:
Thanks :+1:

Issue #571, opened 2015-12-06
Fatal error when create model with --extend option


mzf:
I create model with this command
```
phalcon model pages2 \
  --name=pages  \
  --output=app/core/Models 
  --extends=Base 
  --namespace=Core\Models 
  --force
```

First time model created successfully. 
If i run second time with existed file i get error
```
Fatal error: Class 'Core\Models\Base' not found in Z:\OpenServer\domains\project\client\trunk\app\core\Models\Pages2.php on line 6

Call Stack:
    0.0000     274976   1. {main}() Z:\OpenServer\domains\phalcon\phalcon-devtools-2\vendor\phalcon\devtools\phalcon.php:0
    0.0070     835712   2. Phalcon\Script->run() Z:\OpenServer\domains\phalcon\phalcon-devtools-2\vendor\phalcon\devtools\phalcon.php:82
    0.0070     836056   3. Phalcon\Script->dispatch() Z:\OpenServer\domains\phalcon\phalcon-devtools-2\vendor\phalcon\devtools\scripts\Phalcon\Script.php:152
    0.0080     840168   4. Phalcon\Commands\Builtin\Model->run() Z:\OpenServer\domains\phalcon\phalcon-devtools-2\vendor\phalcon\devtools\scripts\Phalcon\Script.php:127
    0.0100    1100128   5. Phalcon\Builder\Model->build() Z:\OpenServer\domains\phalcon\phalcon-devtools-2\vendor\phalcon\devtools\scripts\Phalcon\Commands\Builtin\Model.php:98
    0.0480    1153176   6. require('Z:\OpenServer\domains\project\client\trunk\app\core\Models\Pages2.php') Z:\OpenServer\domains\phalcon\phalcon-devtools-2\vendor\phalcon\devtools\scripts\Phalcon\Builder\Model.php:339
```

sergeyklay:
What version of Devtools did you use?

web777:
Devtools downloaded as zip 02.12.2015

sergeyklay:
2.0.9 ?
https://github.com/phalcon/phalcon-devtools/archive/v2.0.9.zip

mzf:
Yes. I used this too.

Hodes:
I have changed devtools to be able to find custom classes, refers to #641. Here's in my repo https://github.com/hodes/phalcon-devtools/tree/2.1.x.

alexbusu:
[...]

PR #641, opened 2016-04-04
enables to define a loader file

Hodes:
Now the phalcon-devtools can require a loader file informed on project config: 

I had the same issue as: https://github.com/phalcon/phalcon-devtools/issues/571

With this changes in devtools, I have added to my project config.php the property:

```php
    'devtools' => [
        'loader'      => APP_PATH . '/app/config/loader.php'
    ],
```

and in my project the loader.php is: 

```php 

$loader = new \Phalcon\Loader();

/**
 * We're a registering a set of directories taken from the configuration file
 */
$loader->registerDirs(
    array(
        $config->application->controllersDir,
        $config->application->libraryDir,
        $config->application->pluginsDir,
        $config->application->modelsDir,
//        $config->application->customHelpersDir
    )
)->register();

```
This Loader is used for entire app. A custom one could be made only for devtools.

Now, the devtools is able execute a loader to register any custom classes to use inside Models. 

Thanks.

### DIFF
--- a/scripts/Phalcon/Builder/Model.php
+++ b/scripts/Phalcon/Builder/Model.php
@@ -174,6 +174,10 @@ class Model extends Component
             );
         }
 
+        if (isset($config->devtools->loader)) {
+            require $config->devtools->loader;
+        }
+
         $namespace = '';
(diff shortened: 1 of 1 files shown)


sergeyklay:
Thanks

PR #648, opened 2016-04-13
Volt, overriding callMacro renders error

oligus:
Extending Volt class and overriding callMacro($name, $arguments) renders errors due to signature mismatch:
ErrorException: Declaration of Pow\Mvc\Volt::callMacro() should be compatible with Phalcon\Mvc\View\Engine\Volt::callMacro($name, array $arguments)

Method signature should probably typehint $arguments as array


sergeyklay:
Thanks

oligus:
Thank you too :-)

PR #651, opened 2016-04-18
Update jquery and bootstrap versions, change file line endings.

dschissler:
This is a fairly basic update to the bootstrap and jquery library versions.  Also an extra space was automatically removed from the EOL on many lines.


sergeyklay:
Thanks

PR #652, opened 2016-04-19
added boolean as return type of findFirst() method

aleksandrzen:
I think, that we can fix it in other version of documentation too. What do you think (I've tested this case only in ` Phalcon 2.0.8`)?


sergeyklay:
We don't add changes directly to `master` branch.

PR #653, opened 2016-04-19
added boolean as return type of findFirst() method

aleksandrzen:
I think, that we can fix it in other version of documentation too (I've tested this case only in Phalcon 2.0.8). As I see, changing `ide/stubs` directory will change other cases too, am I right?


sergeyklay:
What other cases?

aleksandrzen:
@sergeyklay I meant that I've tested `findFirst()` method only in `Phalcon 2.0.8`, so if there's other behaviour of this method in other version of framework, maybe this is not correct to change it in `2.0.x` (as I heard, Phalcons' community doesn't follow semantic versioning, so there is possibility that there're different interface in minor version too).

sergeyklay:
First of all need check phpdoc in phalcon repo

PR #686, opened 2016-06-04
Update template DI services to Phalcon 2.1 conventions.

dschissler:
This is a conservative an obvious improvement to new conventions and capabilities.

Later I'd like to make another commit so that the config service requires the config file directly from within the service definition.  This change maybe contentious and not idea so we can look at that separately.

### DIFF
--- a/templates/project/simple/services.php
+++ b/templates/project/simple/services.php
@@ -21,7 +21,9 @@ $di = new FactoryDefault();
 /**
  * The URL component is used to generate all kind of urls in the application
  */
-$di->setShared('url', function () use ($config) {
+$di->setShared('url', function () {
+    $config = $this->getConfig();
+
     $url = new UrlResolver();
     $url->setBaseUri($config->application->baseUri);
 
@@ -31,14 +33,15 @@ $di->setShared('url', function () use ($config) {
 /**
  * Setting up the view component
  */
-$di->setShared('view', function () use ($config) {
+$di->setShared('view', function () {
+    $config = $this->getConfig();
 
     $view = new View();
-
     $view->setViewsDir($config->application->viewsDir);
 
     $view->registerEngines([
-        '.volt' => function ($view, $di) use ($config) {
+        '.volt' => function ($view, $di) {
+            $config = $this->getConfig();
 
             $volt = new VoltEngine($view, $di);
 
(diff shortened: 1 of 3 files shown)


templates/project/simple/services.php
```diff
@@ -21,7 +21,9 @@
 /**
  * The URL component is used to generate all kind of urls in the application
  */
-$di->setShared('url', function () use ($config) {
+$di->setShared('url', function () {
+    $config = $this->getConfig();
```
sergeyklay: `$this->getConfig()` === `$di->getShared('config')`

The Di has `config` service?

dschissler: Oops.

dschissler: Ok, so should I add it?

dschissler: Ok then something like this should be added:

```php
$di->setShared('config', function () use ($config) {
    return $config;
});
```

dschissler: Probably l'll just junk this pull request and to do it differently.  My goal is to have the bootstrap look something like this:

```php
$di = new DI()
require_once('...services.php');

$config = $di->getConfig();
```

and services.php

```php
$di->setShared('config', function () {
    return new Config(require_once(__DIR__ . '/config.php'));
});
```

sergeyklay: 1. more correctly
  ```php
  $di = DI::getDefault();
  ```

2. `__DIR__ ` - You should double check this. I mean you need to make sure it always works the same way if I want to use it as
  * `php public/index.php`
  * `php index.php`
  * `php /full/path/to/the/public/index.php`
  * `php ../public/index.php`
  * `php ../index.php`

dschissler: I don't see how a bootstrap file could be implemented completely independently of the configuration directory.  At some point the bootstrapper needs to know how to find the config directory.

I personally don't like how much stuff is in the main index.php file.  I think that should be very minimal and then requiring the boostrap that resides in the root of the app directory.  Then simply put that bootstrap file must know how to reach the config directory with hard coded paths based off of `__DIR__` and there is really no way around that.  Right?

I don't think that `$di = DI::getDefault();` is needed anywhere other than for use in a static class function.  I really try not to use it and I've found ways around it every time.

dschissler: Ohh, ok about 1) `DI::getDefault()`.  I agree since that fills in default service names if they aren't defined.  I thought that for some reason you meant to sprinkle that around in code instead of just using the then defined `$di`.

dschissler: For 2) __DIR__ should always work the same since its independent of how that file was called and it always gives absolute path.

sergeyklay:
@dschissler

dschissler:
I'm going to abandon this commit and start over with more changes.  I'll get to it some days from now.

Issue #570, opened 2015-12-06
There is no generation of models with annotations. 


mzf:
There is no generation of models with annotations. 
Please add it to create-model action

ITspirit:
+1

EpicJhon:
+1

alexbusu:
This is already in 2.1.x, and starting 3.0.0

PR #692, opened 2016-06-15
Implement models generation with annotations (in 2.1.x)

alexbusu:
This PR will bring the enhancement requested in #570


sergeyklay:
Thanks

PR #707, opened 2016-07-07
Refactor bootstrapping in project templates

dschissler:


dschissler:
Damn, I just forked the repository and it didn't tell me that I already had a repository.  It just took me to my existing repository....  Annoying.  So what to do @sergeyklay?

templates/project/micro/config.php
```diff
@@ -1,7 +1,5 @@
 <?php
 
-defined('APP_PATH') || define('APP_PATH', realpath('.'));
-
```
sergeyklay: @dschissler Why?

dschissler: Why wouldn't it be set already?  I don't feel strongly about that so it can just be added back.

sergeyklay: For example we can use Devtools with Micro app. Devtools starts from `public` dir

dschissler: But I think that the index.php there which is sort of a all-in-one bootstrap already defines it.

sergeyklay: By default Webtools does not use index.php

dschissler: I didn't test it with Webtools.

sergeyklay:
Thanks

PR #708, opened 2016-07-08
Remove old comment about global variables being declared in file.

dschissler:
These variables aren't declared in the services file anymore.


sergeyklay:
Thanks

PR #719, opened 2016-08-02
Update Migrations.php

free6k:
Added support Postrgresql :(

was error:

ERROR: SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "`"
LINE 1: SELECT * FROM `phalcon_migrations` ORDER BY `version` DESC L...


sergeyklay:

PR #720, opened 2016-08-02
Update Migrations.php

free6k:
Added support Postrgresql :(

was error:

ERROR: SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "`"
LINE 1: SELECT * FROM `phalcon_migrations` ORDER BY `version` DESC L...


scripts/Phalcon/Migrations.php
```diff
@@ -516,7 +516,7 @@ public static function getCompletedVersions($options)
         if (isset($options['migrationsInDb']) && (bool)$options['migrationsInDb']) {
             /** @var AdapterInterface $connection */
             $connection = self::$_storage;
-            $completedVersions = $connection->query('SELECT `version` FROM `phalcon_migrations` ORDER BY `version` DESC')->fetchAll();
+            $completedVersions = $connection->query('SELECT version FROM phalcon_migrations ORDER BY version DESC')->fetchAll();
```
sergeyklay: Best way is to use escape methods from Dialect class

free6k: Postgresql don't support this escape, oracle don't support. You write all code for mysql, mysql dont support common SQL standarts...

free6k: Best way for this case it use PDO variables binding, like did it yii2 framework

sergeyklay: Yeah, this is why I suggest to improve this PR. What do you think about this?

free6k: I will make

free6k: ping @sergeyklay, look pull request

free6k: About ```$connection->query('SELECT version FROM phalcon_migrations ORDER BY version DESC')```, i didnt found method for normal queryBuilder for Phalcon.

write ```$connection->query('SELECT '. $connection->escapeString('version') .' .... ``` not good

sergeyklay: https://docs.phalconphp.com/en/latest/api/Phalcon_Mvc_Model_Query_Builder.html

free6k: This query builder working only with model. Here need classical query builder.

```
$qb =  new Builder([
   'from' => self::$migration_log_table,
   'columns' => '*',
   'order' => ['version DESC'],
   'limit' => 1,
]);
echo $qb->getQuery();
```
i have next error:
```
Error: At least one model is required to build the query
```

or you want that i created model for migration table? i think it is wrong

sergeyklay:
Thanks

PR #722, opened 2016-08-02
Add BASE_PATH constant to each template and where it is applicable move cache directory down one level from the apps directory into the base directory.

dschissler:


sergeyklay:
Thanks

Issue #728, opened 2016-08-04
phalcon-devtools/scripts/Phalcon/Web/JQuery.php not updated for new jquery version


silverdr:
You seem to have forgotten to update things there. Hardcoding specific files there doesn't seem like a good idea anyway.

sergeyklay:
@dschissler Could you please take a look?

dschissler:
@silverdr Try the 3.0.x branch.

silverdr:
@dschissler The 3.0.x branch fixes the version numbers but still refers to 'min' file, which is not provided in the resources.

silverdr:
Here the fix should be to replace the static jquery file with a minimised variant in the resources.

sergeyklay:
@silverdr Could you please try 3.0.x now?

silverdr:
The basic functionality seem to work in the 3.0.x branch now. At least I don't see any critical issues like missing CSSs or JSs anymore. Still - in the longer run it would IMHO be better to remove the hardcoded filenames completely and just copy over whatever is present in the respective resources directories. That way resources updates will become far less error prone.

dschissler:
We need to remove the hard coded names and also to not use both local and remote resources for different parts of the tools.  Its just double the maintenance.

dschissler:
So close issue?

silverdr:
> We need to remove the hard coded names and also to not use both local and remote resources for different parts of the tools. Its just double the maintenance.

Precisely, noticed that too. Closing then.

PR #729, opened 2016-08-05
Fixes

dschissler:
When I have time I will move the `APP_PATH` and `BASE_PATH` defines into `index.php` and `launch` for the cli script.  Then I will make it the responsibility of anything that calls into `config.php` to first set these values up.  So for now I'm just expanding on this old work around and when I have time I will improve the design.


sergeyklay:
Thanks

dschissler:
@sergeyklay There is likely an issue with the Volt setup on at least the `modules` project.  I'm fixing that now.

sergeyklay:
What about https://github.com/phalcon/phalcon-devtools/issues/728

dschissler:
I think that this probably fixes that although he was likely referring to a much older branch.  Please test it a little if you have time because I am busy now with other related activities.

dschissler:
The devtools sometimes refers to to a CDN for jquery and other times the local resources.  This should eventually be resolved and the project should use a constant for jquery versions as well.  I won't be doing that work for a while so maybe for Phalcon 3.1.

PR #730, opened 2016-08-05
Improvements

dschissler:
I'm leaving the Volt setup as it is in the `simple` module and the `modules` template is getting some custom work to demonstrate how to change the Volt compiled path.

### DIFF
--- a/templates/project/modules/bootstrap_web.php
+++ b/templates/project/modules/bootstrap_web.php
@@ -45,10 +45,7 @@ try {
      * Register application modules
      */
     $application->registerModules([
-        'frontend' => [
-            'className' => '@@namespace@@\Modules\Frontend\Module',
-            'path' => APP_PATH . '/modules/frontend/Module.php'
-        ],
+        'frontend' => ['className' => '@@namespace@@\Modules\Frontend\Module'],
     ]);
 
     /**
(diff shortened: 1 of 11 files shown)


templates/project/modules/bootstrap_web.php
```diff
@@ -45,10 +45,7 @@
      * Register application modules
      */
     $application->registerModules([
-        'frontend' => [
-            'className' => '@@namespace@@\Modules\Frontend\Module',
-            'path' => APP_PATH . '/modules/frontend/Module.php'
-        ],
```
dschissler: https://github.com/phalcon/cphalcon/blob/3.0.x/phalcon/cli/console.zep#L117

I think not because that key simply doesn't exist and its optional.  If a registered loader knows how to find the module class then everything works well.  I've been using this approach in my private project for a year with no issues.  See the Slacks conversation there today.  I'd like to deprecate this for 4.0 and to only use the loader for finding the path.  The Application classes will still need to know what class to load though.  I propose that we first add a `$modulesName => $className` format for the registerModules method and then towards the end of 3.* series we give a deprecation warning for offering the path there with a full array.

dschissler: You can see that adding the path just acts as a poor mans loader and that the step is completely ignored if the key is missing.

PR #733, opened 2016-08-05
Set Volt service to shared and change its name from 'voltService' to 'voltShared'

dschissler:
This is merely a convention change that I hope will ease some of the difficulties that people have had with the Volt service not working correctly when normal and simple views are combined.

### DIFF
--- a/templates/project/modules/Module_web.php
+++ b/templates/project/modules/Module_web.php
@@ -4,6 +4,7 @@ namespace @@namespace@@\Modules\Frontend;
 use Phalcon\DiInterface;
 use Phalcon\Loader;
 use Phalcon\Mvc\View;
+use Phalcon\Mvc\View\Engine\Php as PhpEngine;
 use Phalcon\Mvc\ModuleDefinitionInterface;
 
 class Module implements ModuleDefinitionInterface
@@ -42,7 +43,7 @@ class Module implements ModuleDefinitionInterface
 
             $view->registerEngines([
                 '.volt'  => 'voltShared',
-                '.phtml' => 'Phalcon\Mvc\View\Engine\Php'
+                '.phtml' => PhpEngine::class
             ]);
 
             return $view;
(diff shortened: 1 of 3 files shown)


dschissler:
I removed the PHP template engine from the `modules` project because there is custom Volt setup.  The `simple` project has both template engines and this is fine because there is no custom setup there.

dschissler:
The Volt service commit doesn't do much in itself but it will make it easier to explain in documentation, projects and examples when a shared and non-shared Volt service is needed.  This becomes necessary when using `Phalcon\Mvc\View\Simple` views to render emails during a web request.

templates/project/modules/Module_web.php
```diff
@@ -41,8 +41,7 @@ public function registerServices(DiInterface $di)
             $view->setViewsDir(__DIR__ . '/views/');
 
             $view->registerEngines([
-                '.volt'  => 'voltService',
-                '.phtml' => 'Phalcon\Mvc\View\Engine\Php'
```
sergeyklay: Return back `.phtml` please

sergeyklay: This is te,plate project. Skeleton. Let user to decide what engine to use

dschissler: Ok.  Should we be copying over the .phtml resources then?  As it is we only move over the `.volt` files in the build script.

templates/project/modules/Module_web.php
```diff
@@ -41,7 +41,8 @@ public function registerServices(DiInterface $di)
             $view->setViewsDir(__DIR__ . '/views/');
 
             $view->registerEngines([
-                '.volt'  => 'voltShared'
+                '.volt'  => 'voltShared',
+                '.phtml' => 'Phalcon\Mvc\View\Engine\Php'
```
sergeyklay: How about? :)

```php
use Phalcon\Mvc\View\Engine\Php as PhpEngine

// ...

$view->registerEngines([
    '.volt'  => 'voltShared',
    '.phtml' => PhpEngine::class
])
```

dschissler: I like that.  I'll have to use it in my projects.

templates/project/modules/Module_web.php
sergeyklay: It seems to me we have to provide just  ability to use this engine. No more

sergeyklay:
Thanks

PR #743, opened 2016-08-09
Fix webtools paths to work with all project templates, in 'micro' app…

dschissler:
… related directories into './app' to be consistent with other projects.


dschissler:
@sergeyklay please test `simple` and `micro` with basic migration tool.

sergeyklay:
Thanks

PR #745, opened 2016-08-10
Fix url service to handle projects that are not in the root web directory

dschissler:


templates/project/micro/services.php
```diff
@@ -25,10 +25,14 @@
  * The URL component is used to generate all kind of urls in the application
  */
 $di->setShared('url', function () {
-    $config = $this->getConfig();
+    $scriptName = $this->getRequest()
+        ->getServer('SCRIPT_NAME');
+
+    $baseUri = preg_replace('/public\/index.php$/', '', $scriptName);
```
sergeyklay: The 
```nginx
root /usr/share/nginx/html; 
```

break it

templates/project/modules/config.ini
```diff
@@ -11,4 +11,3 @@ controllersDir = ../apps/frontend/controllers/
 modelsDir      = ../apps/frontend/models/
 migrationsDir  = ../apps/migrations/
 viewsDir       = ../apps/frontend/views/
-baseUri        = /@@name@@/
```
sergeyklay: Do you suggest disallow user to configure this and just hardcode this rule?

dschissler: Yes.  I think that if anything it should just be allowing the user to configure a prefix or postfix on something.

templates/project/micro/services.php
dschissler: I'm not currently able to test with nginx due to another custom environment.

templates/project/modules/config.ini
sergeyklay: 10 out of 10 frameworks allows user to do it myself. Can you name at least one reason, why hardcoded option it is OK

templates/project/micro/services.php
sergeyklay: Devtools is not desogned for your environment

templates/project/modules/config.ini
dschissler: 1) I said so.

Ok so I'll dump this and give it another go in some days.

PR #746, opened 2016-08-11
Fix in DI services definition

le51:
as in master branch

### DIFF
--- a/scripts/Phalcon/Web/Tools.php
+++ b/scripts/Phalcon/Web/Tools.php
@@ -302,8 +302,7 @@ class Tools
 
             $di->setShared('config', $config);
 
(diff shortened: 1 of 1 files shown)


sergeyklay:
`Phalcon\Di` is now bound to services closures allowing use `Phalcon\Di` as `$this` to access services within them. 

> https://blog.phalconphp.com/post/phalcon-3-0-0-released

le51:
Hi Sergey,

I agree, but as is in 3.0.x branch, it works with php 7.0.8 and doesn't work with php 5.6.11

sergeyklay:
@le51 So, this is Phalcon issue. Could you please open issue in Phalcon repo with script to reproduce (example)?

PR #767, opened 2016-08-24
Ensure BIGINT fields return an integer type for generated models.

daniel18387:
When setting BIGINT field types, the generated models were returning an attribute type of `string`.

## Example

Previously, the `type` and the `@var` were set to `string`.
```
    /**
     *
     * @var string
     * @Primary
     * @Identity
     * @Column(type="string", length=20, nullable=false)
     */
    protected $id;
```
Now, with the changes, they are set to `integer`, which is the expected outcome.
```
    /**
     *
     * @var integer
     * @Primary
     * @Identity
     * @Column(type="integer", length=20, nullable=false)
     */
    protected $id;
```
This has been a big bug for my project as I'm having to update all the getters and setters too. As you can imagine, it's pretty time consuming for big projects.


sergeyklay:

daniel18387:
Ok, no problem.

daniel18387:
@sergeyklay The PR now goes to the `3.0.x` branch.

sergeyklay:
Thanks

PR #770, opened 2016-08-26
fix initial migration timestamp based

### LINKED ISSUE #750: Initial Migration Timestamp Based problem
Hi,

If we create a timestamp based migrations and we try to run the first one, exemple:

```
./phalcon migration run --log-in-db --ts-based
```

We get this error:
```
ERROR: Wrong version number provided
```

Because since there is nothing on database on getCurrentVersion it tries to create a `VersionCollection::createItem(null);` this uses default value `000` that does not match the expected regex.

dagcosta:
fix #750 - this fixes the problem when migration is timestampbased - default value is not according to REGEX requirement


sergeyklay:
Thanks

PR #772, opened 2016-08-28
Fixed possible problems with PostgreSQL-Connections in generated code

hsegnitz:
charset is not a parameter for PostgreSQL Connections and thus switching to PostgreSQL inside an automatically generated project throws exceptions whenever something a database connection is attempted.

this proposed change keeps it more flexible by unsetting the key 'charset' from the adapters config array if the adapter is PostgreSQL


sergeyklay:
Thanks!

PR #776, opened 2016-09-06
Update phalcon.sh

ch3k1:
You need spaces around the comparison operator. 
Problematic code:
```
if [ `echo $0 | grep "bash"`=="bash" ]; then
```

Correct code:

```
if [ `echo $0 | grep "bash"` == "bash" ]; then
```


sergeyklay:
@ch3k1 

ch3k1:
of course

PR #778, opened 2016-09-07
Added bootstrap autoloader

sergeyklay:


alexbusu:
```php
// ./bootstrap/autoload.php file
if (file_exists(PTOOLSPATH . DS . '.phalcon' . DS . 'autoload.php')) {
    require_once PTOOLSPATH . '/.phalcon/autoload.php';
    require_once PTOOLSPATH . DS .  '.phalcon' . DS . 'autoload.php';
  }
}
```
So we now have to put the custom autoloader under vendor dir. No more versioning for the autoloader? :)
Previously, in `./phalcon.php` we had
```php
if (file_exists('.phalcon/autoload.php')) {
    require_once '.phalcon/autoload.php';
}
```
so we were meant to put the custom autoloader inside `PROJECT_PATH/.phalcon` dir.

Is there any particular reason why we should put the custom autoloader under `PTOOLSPATH/.phalcon` dir?

Thank you

sergeyklay:
Hello @alexbusu 

Looks like a bug. Could you please send PR to the 3.0.x branch?

PR #816, opened 2016-09-14
fixed migrations - when running back more than one migration

dagcosta:
HI,

I was having a issue rolling back 2 migrations. So the Scenario:

2 migrations:
Test1:
```
class StatementTypeMigration_1473782767490667 extends Migration
{
    /**
     * Run the migrations
     *
     * @return void
     */
    public function up()
    {
        echo 'UP1' . PHP_EOL;
    }

    /**
     * Reverse the migrations
     *
     * @return void
     */
    public function down()
    {
        echo 'DOWN1' . PHP_EOL;
    }
}
```

and Test2:

```
class StatementTypeMigration_1473782773155414 extends Migration
{
    /**
     * Run the migrations
     *
     * @return void
     */
    public function up()
    {
        echo 'UP2' . PHP_EOL;
    }

    /**
     * Reverse the migrations
     *
     * @return void
     */
    public function down()
    {
        echo 'DOWN2' . PHP_EOL;
    }
}
```

Doing migration run i get UP1 and UP2. Rolling back both, i get DOWN2 and DOWN2. So let's analyse run method.

It was trying to run:
InitialVersion => test2 | VersionItem => test2
InitialVersion => test2 | VersionItem => test1

If we see ModelMigration::migrate method on down it uses InitialVersion to run the down method, that's why it runs test2 down two times.

So my commit... was to make this:
InitialVersion => test2 | VersionItem => test1
InitialVersion => test1 | VersionItem => previous

This way it runs as supposed test2->down() test1->down() and uses morpth from previous if it's present.

Code explanation:
```
if ($initialVersion->getVersion() === $versionItem->getVersion() && ModelMigration::DIRECTION_BACK === $direction) {
    continue;
}
```
Criteria to when Rolling Back to skip from test2 to test2.

```
} elseif ((ModelMigration::DIRECTION_BACK === $direction) && !isset($completedVersions[(string)$initialVersion])) {
    print Color::info('Version ' . (string)$initialVersion . ' was already rolled back');
    $initialVersion = $versionItem;
```
To see if we have or not to rollback we need to check initialVersion and not versionItem since Migrate uses Initial to run the down method. Also increased InitialVersion.

```
if ($initialVersion->getVersion() === $finalVersion->getVersion() && ModelMigration::DIRECTION_BACK === $direction) {
```
Changed criteria to stop, now only stops if initial is the same as final.

```
$migrationsDir . DIRECTORY_SEPARATOR . (ModelMigration::DIRECTION_BACK === $direction ? $initialVersion->getVersion() : $versionItem->getVersion())
```
When down it uses initialVersion to analyse what files to parse since Migrate uses Initial to run the down method.

```
self::removeCurrentVersion($options, (string)$initialVersion);
print Color::success('Version ' . $initialVersion . ' was successfully rolled back');
```
When down, we removes initialVersion and not versionItem, again... since it runs the down on the initial not the versionItem.

Just my 2cents: Same method for up and down, is a bit confusing. I did this commit to solve the issue we were having but i think this should be changed a lot.

### DIFF
--- a/scripts/Phalcon/Migrations.php
+++ b/scripts/Phalcon/Migrations.php
@@ -244,22 +244,30 @@ class Migrations
         // Run migration
         $versionsBetween = VersionCollection::between($initialVersion, $finalVersion, $versionItems);
         foreach ($versionsBetween as $versionItem) {
-            if ((ModelMigration::DIRECTION_FORWARD === $direction) && isset($completedVersions[(string)$versionItem])) {
-                print Color::info('Version ' . (string)$versionItem . ' was already applied');
-                continue;
-            } elseif ((ModelMigration::DIRECTION_BACK === $direction) && !isset($completedVersions[(string)$versionItem])) {
-                print Color::info('Version ' . (string)$versionItem . ' was already rolled back');
+
+            // If we are rolling back, we skip migrating when initialVersion is the same as current
+            if ($initialVersion->getVersion() === $versionItem->getVersion() && ModelMigration::DIRECTION_BACK === $direction) {
                 continue;
             }
 
-            if ($versionItem->getVersion() === $finalVersion->getVersion() && ModelMigration::DIRECTION_BACK === $direction) {
+            if ((ModelMigration::DIRECTION_FORWARD === $direction) && isset($completedVersions[(string)$versionItem])) {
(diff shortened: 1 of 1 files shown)


sergeyklay:
Thanks

PR #835, opened 2016-09-17
Added use

cbichis:
Hi,

Added use to include one other namespace.


sergeyklay:
@cbichis 

PR #843, opened 2016-09-21
Update Snippet.php

radeczo:
Model::find() returns array, but we also can do something like this: Model::find()->update(...) and then we have a warning in IDE.


sergeyklay:
Thanks

PR #857, opened 2016-09-27
bigint and bigserial type fix

radeczo:


sergeyklay:
@radeczo 

radeczo:
Fix is already in 3.0.x

PR #865, opened 2016-09-29
proxy composer bin

nueko:
with this PR, we can just execute phalcon on the command line.


sergeyklay:

nueko:
Did it, but it conflict.

PR #867, opened 2016-09-29
bin proxy

nueko:
with this PR, we can just run `phalcon` on the cli.


nueko:
for windows env reason, we don't use symlink.

sergeyklay:
Thanks

nueko:
You're welcome

PR #902, opened 2016-10-24
Allow 'simple' project template 'url' service to understand non-root project paths.

dschissler:
The issue is that dev tools templates only understand being in the root directory of a web space or vhost and this is never the case in my projects.  One downside of this pull request is that now the ini config doesn't match the normal one but I don't see how this can be done with an ini file.  Its still no worse for ini then it was so it will just be a different flavor.

* This should handle Windows and Unix path separators.
* This only changes the `simple` template and later I will add the `modules` fix as well (but I'm having much more serious issues with `modules` at the moment).

### DIFF
--- a/templates/project/simple/config.php
+++ b/templates/project/simple/config.php
@@ -24,6 +24,10 @@ return new \Phalcon\Config([
         'pluginsDir'     => APP_PATH . '/plugins/',
         'libraryDir'     => APP_PATH . '/library/',
         'cacheDir'       => BASE_PATH . '/cache/',
+
+        // This allows the baseUri to be understand project paths that are not in the root directory
+        // of the webpspace.  This will break if the public/index.php entry point is moved or
+        // possibly if the web server rewrite rules are changed. This can also be set to a static path.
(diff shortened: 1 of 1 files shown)


sergeyklay:
* Ini config, Json config, Xml config?

sergeyklay:
Do you really think it cannot be done individually?

dschissler:
I wasn't aware that there are other static configs than the Ini one.

It can be done individually if I understand your question but consider the following realistic scenario:

Joe just setup his Windows WAMP environment from an exe and he barely knows what he is doing.  He creates a directory `phalcon_projects` off of the webspace and then he creates a project named `test`.  Well now all internal URLs are broken because the "url" service doesn't consider the first path `phalcon_projects/`.

Not to mention that it is a pain in the ass in my little projects and it creates unnecessary steps.  Someone could always go in and hard code a path in the config if they want to do so later.

sergeyklay:
I suggest to leave a detailed comment like "why is it necessary and when it may hinder" in `templates/project/simple/config.php`

dschissler:
What are the situations?  Maybe if `public/index.php` is moved?  What else?

sergeyklay:
Thanks

PR #904, opened 2016-11-03
Fixes and cleanups for 'modules' project

dschissler:
I added the 'url' fix fom 'simple' project.

Another issue was was due to some of the `modules` code being copied from `simple` where `Module` was being replaced with `Controllers`.  I changed it to match to the end of the string because the module classes have two "Module" strings in them, once for "Modules" and then another time at the end for the class name itself.

I tested a dispatcher forward as well as the following paths and they all work:

* /projects_container/test_modules/
* /projects_container/test_modules/index
* /projects_container/test_modules/index/index
* /projects_container/test_modules/frontend
* /projects_container/test_modules/frontend/index
* /projects_container/test_modules/frontend/index/index
* /projects_container/test_modules/frontend/test
* /projects_container/test_modules/frontend/test/other

### DIFF
--- a/templates/project/modules/routes.php
+++ b/templates/project/modules/routes.php
@@ -1,9 +1,9 @@
 <?php
 
-$router = $di->get("router");
+$router = $di->getRouter();
 
 foreach ($application->getModules() as $key => $module) {
-    $namespace = str_replace('Module','Controllers', $module["className"]);
+    $namespace = preg_replace('/Module$/', 'Controllers', $module["className"]);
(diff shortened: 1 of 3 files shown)


templates/project/modules/routes.php
```diff
@@ -26,5 +26,3 @@
         'params' => 3
     ]);
 }
-
-$di->set("router", $router);
```
sergeyklay: Why?

Issue #906, opened 2016-11-11
Modules Template for Routers


tsotsos:
Creating a modules application you may notice that automatic route loop produces wrong namespace, since it replaces both 'Module' and 'Modules' with 'Controller' string. As a result routers are useless since they not match to actual namespaces.

tsotsos:
Already solved at branch phalcon:3.0.x

PR #907, opened 2016-11-11
Solution for github issue #906 "Modules Template for Routers"

tsotsos:
Creating a modules application you may notice that automatic route loop produces wrong namespace, since it replaces both 'Module' and 'Modules' with 'Controller' string. As a result routers are useless since they not match to actual namespaces.

Signed-off-by: Georgios Tsotsos <tsotsos@gmail.com>

### DIFF
--- a/templates/project/modules/routes.php
+++ b/templates/project/modules/routes.php
@@ -3,7 +3,7 @@
 $router = $di->get("router");
 
 foreach ($application->getModules() as $key => $module) {
-    $namespace = str_replace('Module','Controllers', $module["className"]);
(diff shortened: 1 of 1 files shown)


sergeyklay:
@tsotsos  Thanks you for contributing.
We don't add changes directly to stable branches or `master`.
I've changed the base branch to `phalcon:3.0.x` from `phalcon:master` just now.
Could you please rebase now? This PR has conflicts with the `phalcon:3.0.x` branch that must be resolved.

tsotsos:
Thank you @sergeyklay , just checked branch phalcon:3.0.x relevant file and its already resolved. I close this PR and the issue #906

PR #914, opened 2016-11-22
Update IncrementalItem.php

nikonm:
Fixed incremental type(for compatible with 2.x versions)


sergeyklay:

PR #917, opened 2016-11-23
Migration csv export

lenrsmith:
The data export functionality was failing for me because of newline characters and pipe delimiters within some of the columns. Not sure if that this is the best way to handle things, but it works for me now, so I wanted to share.


sergeyklay:

PR #918, opened 2016-11-23
Fixed error when TimestampedItem::_construct is called with '000' ver…

lenrsmith:
…sion number

Trying this again. Thanks for pointing me in the right direction.


sergeyklay:
Thanks!

PR #919, opened 2016-11-23
Use csv for data export

lenrsmith:
This is just a suggestion, but I ran into some very annoying issues trying to import a large dataset with newlines inside text and memo fields. There are also some typo fixes (one in a comment and another in Phalcon\Version\VersionAwareTrait that was breaking migrations)


sergeyklay:
Thanks

Issue #775, opened 2016-09-06
How to generate a migration of multiple tables?


DannyFeliz:
I need to generate a migration of two tables with this command 
`phalcon migration generate --table=tableA tableB` 
but the dev-tool only takes the last specified table, there is a way to get this working?

lenrsmith:
@sergeyklay I'm thinking about ways to handle this as it would be nice to have for one of my projects. It seems we are grabbing options with $argv and could make use of some more robust options handling. Any thoughts here? I can make this work for CLI, but I don't want to break the Web Tools.

sergeyklay:
It seems to me we can check for string/array type of argument to handle multiple tables

lenrsmith:
I'll dig into this a bit. I'm just getting my head into how the dev-tools work, so bear with me.

PR #922, opened 2016-11-25
Enable multiple, comma delimited table names for --table option

lenrsmith:
This should be able to handle multiple tables listed with the `--table` option to address Issue https://github.com/phalcon/phalcon-devtools/issues/775

### DIFF
--- a/scripts/Phalcon/Migrations.php
+++ b/scripts/Phalcon/Migrations.php
@@ -149,12 +149,17 @@ class Migrations
                     ) || $wasMigrated;
             }
         } else {
-            $migration = ModelMigration::generate($versionItem, $tableName, $exportData);
-            $tableFile = $migrationPath . DIRECTORY_SEPARATOR . $tableName . '.php';
-            $wasMigrated = !!file_put_contents(
-                $tableFile,
-                '<?php ' . PHP_EOL . PHP_EOL . $migration
-            );
+            $tables = explode(',', $tableName);
+            if(is_array($tables)) {
+                foreach($tables as $table) {
+                    $migration = ModelMigration::generate($versionItem, $table, $exportData);
+                    $tableFile = $migrationPath . DIRECTORY_SEPARATOR . $table . '.php';
+                    $wasMigrated = !!file_put_contents(
+                        $tableFile,
+                        '<?php ' . PHP_EOL . PHP_EOL . $migration
+                    );
+                }
+            }
         }
 
         if (self::isConsole() && $wasMigrated) {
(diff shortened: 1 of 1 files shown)


scripts/Phalcon/Migrations.php
```diff
@@ -149,12 +149,17 @@ public static function generate(array $options)
...
             }
         } else {
-            $migration = ModelMigration::generate($versionItem, $tableName, $exportData);
-            $tableFile = $migrationPath . DIRECTORY_SEPARATOR . $tableName . '.php';
-            $wasMigrated = !!file_put_contents(
-                $tableFile,
-                '<?php ' . PHP_EOL . PHP_EOL . $migration
-            );
+            $tables = explode(',', $tableName);
+            if(is_array($tables)) {
```
sergeyklay: The `explode(',', $tableName);` always will return array

scripts/Phalcon/Migrations.php
```diff
@@ -149,12 +149,17 @@ public static function generate(array $options)
...
+                foreach($tables as $table) {
```
sergeyklay: Could you please use common style? ` foreach(` =>  ` foreach (`

scripts/Phalcon/Migrations.php
```diff
@@ -149,12 +149,17 @@ public static function generate(array $options)
...
+                    $migration = ModelMigration::generate($versionItem, $table, $exportData);
+                    $tableFile = $migrationPath . DIRECTORY_SEPARATOR . $table . '.php';
+                    $wasMigrated = !!file_put_contents(
```
sergeyklay: Could you avoid to use `!!`? The `file_put_contents` function returns **the number of bytes** that were written to the file, or **FALSE** on failure.

lenrsmith: This is what was in the code, originally. I did not change it, but I will fix it.

scripts/Phalcon/Migrations.php
lenrsmith: Certainly.

scripts/Phalcon/Migrations.php
lenrsmith: True, I was thinking it could return *false*, but that is only if you supply "" as the delimiter. I'll fix that that.

sergeyklay:
Thanks

wmathes:
@sergeyklay Thanks for the feature! Please close #775 :) it refers to this and is still open :)

DannyFeliz:
Thanks!

Issue #238, opened 2014-08-07
Migration adds fixed value of referencedSchema in references


siciarek:
Migration adds fixed value of referencedSchema in references
it forces developer to use the same table name in dev, test and prod environments

Possible solution:

``` php
protected static $_databaseConfig; // line 53

$referenceDefinition[] = "'referencedSchema' => " . 'self::$_databaseConfig->dbname'; // line 259
```

7thcubic:
this is a known bug but its not because of migration. Its because of phalcon. I made a fix for 2.0
Dev tools does correctly generate it. Its just the creation part.

https://github.com/7thcubic/cphalcon/commit/2fbfe0a9cbb45bc843960b7511cfc6f1ee955632

sergeyklay:
Fixed in the `3.2.x branch`. Feel free to open a new issue if the problem appears again. Thank you for contributing.

PR #946, opened 2016-12-21
fix db referencing problem

philhsu:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/phalcon-devtools/issues/238

This pull request affects the following components: **(please check boxes)**

- [ ] Core
- [ ] WebTools
- [x] Migrations
- [ ] Models
- [ ] Scaffold
- [ ] Documentation
- [ ] IDE
- [ ] MySQL
- [ ] PostgreSQL
- [ ] Oracle
- [ ] SQLite
- [ ] Testing
- [ ] Code Quality
- [ ] Templating

- [x] I have read and understood the [Contributing Guidelines][:contrib:]?
- [x] I have checked that another pull request for this purpose does not exist.
- [ ] I wrote some tests for this PR.

Small description of change:

This bug occurs when the development db name is different from the production db name. When the migration tool generates the migration script from the development environment, the references contains the development db name (referencedSchema). When doing the migration in production server, the migration tool still reads referencedSchema even the db config is set to production because the db name in referencedSchema is hardcoded within the script. By removing referencedSchema from generating the migration script will fix this problem. Since this migration tool does not support multiple db schemas currently, db name should be default to the config db name instead of reading from the script.

Thanks

### DIFF
--- a/scripts/Phalcon/Mvc/Model/Migration.php
+++ b/scripts/Phalcon/Mvc/Model/Migration.php
@@ -332,7 +332,6 @@ class Migration
             }
 
             $referenceDefinition = [];
-            $referenceDefinition[] = "'referencedSchema' => '".$dbReference->getReferencedSchema()."'";
             $referenceDefinition[] = "'referencedTable' => '".$dbReference->getReferencedTable()."'";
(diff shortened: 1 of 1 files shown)


sergeyklay:
Thanks

PR #1018, opened 2017-04-28
Fix:

arthurcn:
1."phalcon all-models --doc" could't generate document for model-class.
2. PhpStrom warning: Return value type is not compatible with declared.

Hello!

* Type: bug fix | code quality
* Link to issue:

Small description of change:

1. "phalcon all-models --doc" will generate document for class. like this:

`	<?php`
`	/**`
`	 * Robots`
`	 * @date 2017-04-29 05:30:34`
`	 */`
`	class Robots`
`	{`
`	}`

2. phpstrom warning: Return value type is not compatible with declared.

`	public static function find($parameters = null)`
`	{`
`		//PhpStrom warning: Return value type is not compatible with declared.`
`		return parent::find($parameters);`
`	}`

Thanks

### DIFF
--- a/scripts/Phalcon/Builder/AllModels.php
+++ b/scripts/Phalcon/Builder/AllModels.php
@@ -200,6 +200,7 @@ class AllModels extends Component
                     'belongsTo' => $belongsToModel,
                     'foreignKeys' => $foreignKeysModel,
                     'genSettersGetters' => $genSettersGetters,
+                    'genDocMethods' => $this->options->get('genDocMethods'),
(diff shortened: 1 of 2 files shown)


arthurcn:
1."phalcon all-models --doc" could't generate document for model-class.
2. PhpStrom warning: Return value type is not compatible with declared.

sergeyklay:
We don't add changes directly to stable branches or `master`. I've changed the base branch to `3.2.x` from `3.1.x`.

Thank you

PR #1039, opened 2017-07-07
directly link js_ie files to avoid slowing down the request

panwenbin:
Hello!

* Type: suggest , 
* Link to issue: 

Small description of change:

I found webtools.php slowed down by remote request js_ie related js files, and it also requires openssl opened, or the page will be blank.   
I think it is not necessary to combine and compress the few used js files which are remote files.   

Thanks

### DIFF
--- a/scripts/Phalcon/Mvc/Controller/Base.php
+++ b/scripts/Phalcon/Mvc/Controller/Base.php
@@ -108,12 +108,8 @@ abstract class Base extends Controller
 
         $this->assets
             ->collection('js_ie')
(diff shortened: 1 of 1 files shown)


sergeyklay:
1. We don't add changes directly to stable branches or `master`
2. Did you try the 3.2.x branch? It seems to me the assets-related issue has been fixed in the  3.2.x branch

Cc: @sergeysviridenko

panwenbin:
the same in 3.2.x branch

PR #1042, opened 2017-07-08
directly link js_ie files to avoid slowing down the request

panwenbin:
Hello!

* Type: suggest 
* Link to issue:

Small description of change:
### I have tried 3.2.x branch, it has the same issue. This is based on 3.2.x branch.

I found webtools.php slowed down by remote request js_ie related js files, and it also requires openssl opened, or the page will be blank.
I think it is not necessary to combine and compress the few used js files which are remote files.

Thanks


sergeyklay:
Thank you

Issue #663, opened 2016-05-10
model generation when hyphen's are in table name or column name


tempcke:
using 2.1.0
I realize it seems stupid to put a - in a table name or column name but I did not create the schema :P

If a table name has a hyphen in it then the generated class name has a hyphen in it which is invalid.
same with column's containing a hyphen .. the resulting properties, getters and setter methods all end up with hyphen's in them.

sergeyklay:
@tempcke Could you please provide a small script to reproduce?

tempcke:
@sergeyklay 
First I made a mistake, I'm using 2.0.10 not 2.1.0 :)
## sql
```sql
# mysql
CREATE TABLE IF NOT EXISTS `foo-bar` (
  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
  `some-col` varchar(20) NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB;
INSERT INTO `foo-bar` (`id`, `some-col`) VALUES (1, 'a');
```
## All models
```
$ phalcon all-models --get-set
```
**Result:**
```php
# app/models/Foo-Bar.php
class Foo-bar extends \Phalcon\Mvc\Model {
    protected $some-col;

    /**
     * Method to set the value of field some-col
     */
    public function setSome-col($some-col)  {
        $this->some-col = $some-col;
        return $this;
    }

    /**
     * Returns the value of field some-col
     */
    public function getSome-col()   {
        return $this->some-col;
    }

    /**
     * Returns table name mapped in the model.
     */
    public function getSource()   {
        return 'foo-bar';
    }
}
```

## one model
```
$ phalcon model --name foo-bar --get-set
```
**Result:**
```php
class FooBar extends \Phalcon\Mvc\Model {
    protected $some-col;

    /**
     * Method to set the value of field some-col
     */
    public function setSome-col($some-col)   {
        $this->some-col = $some-col;
        return $this;
    }

    /**
     * Returns the value of field some-col
     */
    public function getSome-col()    {
        return $this->some-col;
    }

    public function initialize()    {
        $this->setSource("foo-bar");
    }

    public function getSource()    {
        return 'foo-bar';
    }
}
```
So when doing one model with `phalcon model` it removes the - from the filename and classname but leaves it in the property and method names.

sergeyklay:
@tempcke  Thanks. I'll try to sort out as soon as I can

tempcke:
[...]

PR #1046, opened 2017-07-10
Create model with hyphen in table name

sergeysviridenko:
Hello!

* Type: bug fix
* Link to issue: #663

**In raising this pull request, I confirm the following :**

Small description of change:
For proper creating model when table name has hyphen

Thanks

### DIFF
--- a/scripts/Phalcon/Builder/Model.php
+++ b/scripts/Phalcon/Builder/Model.php
@@ -28,6 +28,7 @@ use Phalcon\Validation;
 use Phalcon\Generator\Snippet;
 use Phalcon\Db\ReferenceInterface;
 use Phalcon\Validation\Validator\Email as EmailValidator;
+use Phalcon\Text;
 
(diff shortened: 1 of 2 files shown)


scripts/Phalcon/Builder/Model.php
```diff
@@ -75,11 +76,11 @@ public function __construct(array $options)
         }
 
         if (!isset($options['className'])) {
-            $options['className'] = Utils::camelize($options['name']);
+            $options['className'] = Text::camelize(Utils::camelize($options['name']));
```
sergeyklay: Can we extend `Utils::camelize` or use `Text::camelize` only ? It is bad idea to use `camelize(camelize(...))`

sergeysviridenko: I've changed.

sergeyklay:
Thank you

Issue #1054, opened 2017-07-23
Prompt the user to create a .phalcon directory if missing from the project


ninjapanzer:
Example:
`$ phalcon migration new_mgration .`

will result in:
```
Phalcon DevTools (3.2.0)

  Error: This command should be invoked inside a Phalcon project directory.
```
The unfortunate part of this is we are not exposing to the use what the devtools are upset about.
This command should either tell the user to create a `.phalcon` directory in the project or ask them to confirm and generate it for them.

sergeyklay:
Fixed in the `3.2.x branch`. Feel free to open a new issue if the problem appears again. Thank you for contributing.

PR #1055, opened 2017-07-24
Make phalcon self-healing when executing an external false command when .phalcon folder is missing

ninjapanzer:
Hello!

* Type: **new feature**
* Link to issue: #1054

Small description of change:

When the exception for a missing .phalcon folder is raised the user is prompted and the missing folder is created automatically. Once the folder is created the command is re-executed.

I have added an interface for "Self-Healing Exceptions" so this can be propagated to other locations where cli execution stops but could be resolved.

Thanks

screen shot 2017-07-23 at 10 20 32 pm


sergeyklay:
Thank you!

Issue #203, opened 2014-04-03
Controller Bug when using getters and setters


garethwi:
There are two places in the controllers where direct access to the properties is being attempted when the get-set option has been chosen:

editAction

```
$this->view->id = $model->id;
```

should be

```
$this->view->id = $model->getId();
```

And in saveAction

```
"params" => array($region->id)
```

should be 

```
"params" => array($region->getId())
```

sergeyklay:
Fixed in the `3.2.x branch`. Feel free to open a new issue if the problem appears again. Thank you for contributing.

PR #1062, opened 2017-07-28
Bug fix controller when use getters and setters

sergeysviridenko:
Hello!

* Type: bug fix
* Link to issue: #203 

**In raising this pull request, I confirm the following:**

Small description of change:
Fixed problem in controller when using command `phalcon scaffold --get-set --table-name=foo-bar`. Generated controller contains without `get` methods.

Thanks


sergeyklay:
Thank you

Issue #215, opened 2014-05-04
Migration showing sql profiles


nazwa:
Hello everyone!

I was just thinking... every time migrations are generated or run, all sql queries are outputted to console. Do we really need that? For me it's just clutter, but do you think those profiles are actually useful to anyone?

rianorie:
perhaps an --output flag, something to think about.

sergeysviridenko:
@nazwa I suppose it useful because I can control process generate/run migration and I can see all errors and all SQL queries which executed. But i agree we can wright SQL queries in `log` file or screen. Is this your suggestion?

sergeysviridenko:
I guess we can close this issue.

sergeysviridenko:
I guess we can close this issue.

PR #1103, opened 2017-09-11
Added option --dry --verbose to migration command

sergeysviridenko:
Hello!

* Type: new feature
* Link to issue: #215

**In raising this pull request, I confirm the following:**

- [x] I wrote some tests for this PR

Small description of change:
Added option `--dry` and `--verbose` option for `migration` command. 

Thanks

### DIFF
--- a/scripts/Phalcon/Commands/Builtin/Migration.php
+++ b/scripts/Phalcon/Commands/Builtin/Migration.php
@@ -4,7 +4,7 @@
   +------------------------------------------------------------------------+
   | Phalcon Developer Tools                                                |
   +------------------------------------------------------------------------+
-  | Copyright (c) 2011-2016 Phalcon Team (https://www.phalconphp.com)      |
+  | Copyright (c) 2011-2017 Phalcon Team (https://www.phalconphp.com)      |
   +------------------------------------------------------------------------+
   | This source file is subject to the New BSD License that is bundled     |
   | with this package in the file LICENSE.txt.                             |
@@ -54,6 +54,8 @@ class Migration extends Command
             'force'             => 'Forces to overwrite existing migrations',
             'ts-based'          => 'Timestamp based migration version',
             'log-in-db'         => 'Keep migrations log in the database table rather than in file',
+            'dry'               => 'Attempt requested operation without making changes to system (Generating only)',
+            'verbose'           => 'Output of debugging information during operation (Running only)',
             'no-auto-increment' => 'Disable auto increment (Generating only)',
             'help'              => 'Shows this help [optional]',
         ];
@@ -111,23 +113,21 @@ class Migration extends Command
         }
 
         $tableName = $this->isReceivedOption('table') ? $this->getOption('table') : '@';
-        $descr = $this->getOption('descr');
-        $exportData = $this->getOption('data');
         $action = $this->getOption(['action', 1]);
-        $version = $this->getOption('version');
 
         switch ($action) {
             case 'generate':
                 Migrations::generate([
                     'directory'       => $path,
                     'tableName'       => $tableName,
-                    'exportData'      => $exportData,
+                    'exportData'      => $this->getOption('data'),
                     'migrationsDir'   => $migrationsDir,
-                    'version'         => $version,
+                    'version'         => $this->getOption('version'),
                     'force'           => $this->isReceivedOption('force'),
                     'noAutoIncrement' => $this->isReceivedOption('no-auto-increment'),
                     'config'          => $config,
-                    'descr'           => $descr,
+                    'descr'           => $this->getOption('descr'),
+                    'verbose'         => $this->isReceivedOption('dry'),
                 ]);
                 break;
             case 'run':
@@ -138,8 +138,9 @@ class Migration extends Command
                     'force'          => $this->isReceivedOption('force'),
                     'tsBased'        => $migrationsTsBased,
                     'config'         => $config,
-                    'version'        => $version,
+                    'version'        => $this->getOption('version'),
                     'migrationsInDb' => $migrationsInDb,
+                    'verbose'        => $this->isReceivedOption('verbose'),
                 ]);
                 break;
             case 'list':
@@ -150,7 +151,7 @@ class Migration extends Command
                     'force'          => $this->isReceivedOption('force'),
                     'tsBased'        => $migrationsTsBased,
                     'config'         => $config,
-                    'version'        => $version,
+                    'version'        => $this->getOption('version'),
                     'migrationsInDb' => $migrationsInDb,
                 ]);
                 break;
--- a/scripts/Phalcon/Migrations.php
+++ b/scripts/Phalcon/Migrations.php
@@ -34,6 +34,7 @@ use Phalcon\Mvc\Model\Exception as ModelException;
 use Phalcon\Mvc\Model\Migration as ModelMigration;
 use Phalcon\Version\IncrementalItem as IncrementalVersion;
 use Phalcon\Version\ItemCollection as VersionCollection;
+use Phalcon\Console\OptionStack;
 
 /**
  * Migrations Class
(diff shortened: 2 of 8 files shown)


scripts/Phalcon/Mvc/Model/Migration.php
```diff
@@ -128,20 +136,13 @@ public static function setup($database)
...
-                function ($event, $connection) use ($profiler) {
-                    if ($event->getType() == 'beforeQuery') {
-                        $profiler->startProfile($connection->getSQLStatement());
-                    }
-                    if ($event->getType() == 'afterQuery') {
-                        $profiler->stopProfile();
-                    }
-                }
+                $dbListener
             );
```
sergeyklay:
```php
$eventsManager->attach('db', function () {
    return new DbProfilerListener();
});
```

scripts/Phalcon/Mvc/Model/Migration.php
```diff
@@ -86,16 +88,22 @@ class Migration
...
      * @throws \Phalcon\Db\Exception
      */
-    public static function setup($database)
+    public static function setup($database, array $settings = null)
     {
         if (!isset($database->adapter)) {
             throw new DbException('Unspecified database Adapter in your configuration!');
         }
 
+        $showMode = false;
```
sergeyklay: profile, verbose?

scripts/Phalcon/Mvc/Model/Migration.php
```diff
@@ -128,20 +136,13 @@ public static function setup($database)
             self::$_connection->setDialect(new PostgresqlExtended);
         }
 
-        if (Migrations::isConsole()) {
-            $profiler = new Profiler();
+        if (Migrations::isConsole() && $showMode) {
```
sergeyklay:
```php
if (!Migrations::isConsole() || !$verbose) {
    return;
}
```

scripts/Phalcon/Mvc/Model/Migration.php
```diff
@@ -38,6 +38,8 @@
 use Phalcon\Db\Dialect\PostgresqlExtended;
 use Phalcon\Db\Adapter\Pdo\PostgresqlExtended as AdapterPostgresqlExtended;
 use Phalcon\Utils\Nullify;
+use Phalcon\Logger\Adapter\File as LoggerFile;
```
sergeyklay: why?

scripts/Phalcon/Migrations.php
```diff
@@ -223,7 +233,9 @@ public static function run(array $options)
             $finalVersion = VersionCollection::maximum($versionItems);
         }
 
-        ModelMigration::setup($config->database);
+        $settings['showEvent'] = $verboseMode;
```
sergeyklay: showEvent  => verbode

scripts/Phalcon/Migrations.php
```diff
@@ -82,6 +82,7 @@ public static function generate(array $options)
         $config = $options['config'];
         $descr = isset($options['descr']) ? $options['descr'] : null;
         $noAutoIncrement = isset($options['noAutoIncrement']) ? $options['noAutoIncrement'] : null;
+        $dryMode = isset($options['dry']) ? $options['dry'] : false;
```
sergeyklay: I propose to move this code (or create new one) to the separated trait: https://github.com/phalcon/phalcon-devtools/blob/416356408ab53fb69de8f23af153391c448e0af9/scripts/Phalcon/Commands/Command.php#L428-L441

And use it here

scripts/Phalcon/Listeners/DbListener.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace Phalcon\Listeners;
```
sergeyklay: Add PHPDoc and License agreement

scripts/Phalcon/Commands/Builtin/Migration.php
```diff
@@ -54,6 +54,8 @@ public function getPossibleParams()
             'force'             => 'Forces to overwrite existing migrations',
             'ts-based'          => 'Timestamp based migration version',
             'log-in-db'         => 'Keep migrations log in the database table rather than in file',
+            'dry'               => 'Mode which should not modify anything',
+            'verbose'           => 'Shows details about the results of running',
```
sergeyklay: Enable verbose mode

sergeyklay: As I can understand from this help we can use `--dry` and `--verbose` together. Right?

scripts/Phalcon/Migrations.php
sergeyklay: $settings undefined here

scripts/Phalcon/Utils/ReceivedOptionTrait.php
```diff
@@ -0,0 +1,57 @@
...
+namespace Phalcon\Utils;
+
+/**
+ * \Phalcon\Utils\ReceivedOption
+ *
+ * @package   Phalcon\Utils
+ * @copyright Copyright (c) 2011-2017 Phalcon Team (team@phalconphp.com)
+ * @license   New BSD License
+ */
+trait ReceivedOptionTrait
```
sergeyklay: Could we rename this to `OptionParserTrait`?

scripts/Phalcon/Utils/ReceivedOptionTrait.php
```diff
@@ -0,0 +1,57 @@
...
+{
+    /**
+     * Parameters received by the script.
+     * @var array
+     */
+    protected static $_parameters = [];
```
sergeyklay: `$_parameters = [];` => `$_parameters = [];`

scripts/Phalcon/Utils/ReceivedOptionTrait.php
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Indicates whether the script was a particular option.
+     *
+     * @param  string  $option
+     * @return boolean
+     */
+    public static function isReceivedOption($option)
```
sergeyklay: Wy we need static methods at all?

scripts/Phalcon/Migrations.php
```diff
@@ -581,4 +565,42 @@ public static function getCompletedVersions($options)
...
         return array_flip($completedVersions);
     }
+
+    /**
+     * Get migration version for generating migartion
+     *
+     * @param array $options Applications options
+     * @return IncrementalVersion
+     */
+    protected static function getVersionNameGeneratingMigration()
```
sergeyklay: Why do we need static access here?

scripts/Phalcon/Migrations.php
```diff
@@ -586,4 +564,42 @@ public static function getCompletedVersions($options)
 
         return array_flip($completedVersions);
     }
+
+    /**
+     * Get migration version for generating migartion
```
sergeyklay: **Get** <strike>migration</strike> **version name** <strike>for</strike> **to generat**<strike>ing</strike>**e** <strike>migartion</strike> **migration**

scripts/Phalcon/Commands/Builtin/Migration.php
```diff
@@ -54,6 +54,8 @@ public function getPossibleParams()
             'force'             => 'Forces to overwrite existing migrations',
             'ts-based'          => 'Timestamp based migration version',
             'log-in-db'         => 'Keep migrations log in the database table rather than in file',
+            'dry'               => 'Mode which should not modify anything for generate migration',
+            'verbose'           => 'Enable verbose mode of running migration',
```
sergeyklay: @ninjapanzer @niden could you please review

ninjapanzer: What migration commands can I do a dryrun on?
Generate / Migrate
I think it might be safer to use the term 'operation' instead of migration to avoid having to specify if you are running or generating migrations.

```
'dry' => 'Attempt requested operation without making changes to system',
'verbose'  => 'Output of debugging information during operation'
```

sergeysviridenko: @ninjapanzer Both of command relate to different operation. And there I specified operation in description. `dry` - `...for generate migration` and `verbose` - `of running migration`.
`'Attempt requested operation without making changes to system'` - this description not obviously relate to generate option. I guess description must have obviously indicate to which option it relate.

sergeyklay: But `which should not modify anything` sounds like a joke for me, e.g. "Should not modify, but from time to time can" hahahaha. There is big difference between "SHOULD NOT" and "MUST NOT". So where is "obvious"? Could you elaborate on that :)

scripts/Phalcon/Commands/Builtin/Migration.php
```diff
@@ -111,23 +113,21 @@ public function run(array $parameters)
...
+                    'exportData'      => $this->getOption('data'),
                     'migrationsDir'   => $migrationsDir,
-                    'version'         => $version,
+                    'version'         => $this->getOption('version'),
                     'force'           => $this->isReceivedOption('force'),
                     'noAutoIncrement' => $this->isReceivedOption('no-auto-increment'),
                     'config'          => $config,
-                    'descr'           => $descr,
+                    'descr'           => $this->getOption('descr'),
+                    'verbose'         => $this->isReceivedOption('dry'),
```
ninjapanzer: should this be `'dry' => $this->isReceivedOption('dry')`

scripts/Phalcon/Commands/Builtin/Migration.php
sergeysviridenko: @sergeyklay yes, you're right. I'd like to say that we have to obviously relate option and description. I think option must contain indicator like `generate` or `run`. But I agree the descriptions not so good and obviously they should be changed. I propose:
`Attempt generate migration without making changes to system` - for generating mode
`Output of debugging information during operation of running migration` - for running mode

ninjapanzer: Agree, replace `should` be replaced with `will not` or `does not`.

Just from a usability side of things dry-run should be applicable to either command.
aka:
`phalcon generate --dry --verbose`
`phalcon run --dry --verbose`

 I may want to `dry`-run a migration to see if any errors would occur. I don't know how this would implement but possibly wrapping all the migrations in a transaction and performing a rollback when complete would be enough. 

The same is true for `verbose` which should globally enable trace messages within the command.

Since these options could logically apply to either command I would like to see this.
If a command cannot/will not support `dry` I would like it to warn me so I don't think it will work just in case I didn't read the description.
On the other hand I suspect that `verbose` should be available for any command and its description should not be specific to a command like `running migration`

Some thoughts for you to ponder

scripts/Phalcon/Commands/Builtin/Migration.php
sergeysviridenko: @ninjapanzer there are two modes. They do similar work but there is a different. `Dry` mode using to generate migration, if turf on it - migration won't be created but user will see all event on the screen. `verbose` mode using if user would like to see all event on time workflow running migration. This a reason why there using two option.

scripts/Phalcon/Commands/Builtin/Migration.php
sergeysviridenko: @ninjapanzer Ok I will replace it.
But roll back doesn't work with adding or deleting column for example. So `dry` mode isn't possible in running migration. for unification command we can use `verbose` for both of run and generate. This option will either show or hide event in process of operation. How do you think  @sergeyklay

ninjapanzer: How about we add a `generate-dry-run` command instead of an option that could be interpreted as shared between both commands?

sergeyklay: I propose to leave --dry and --verbose. We may want to increase verbosity for other commands which doesn't have --dry mode. Actually --verbose just logger (to STDOUT) in other hand --dry may prevent normal command execution. We just need add a notice for Migration-related command that --dry does not work (yet) with structural changes so we can't use transactions to do --dry magic

sergeyklay: > does not work (yet) with structural changes

This is why 

> Mode which should not modify anything for generate migration

totally incorrect

sergeysviridenko: @ninjapanzer just only `verbose`. because `dry` can't be added.
In sum I add just command `verbose`. it will either show or hide all operation events. OK?
What do you think about @sergeyklay?

sergeyklay:
Thank you

PR #1109, opened 2017-09-24
Fix: belongsTo with namespace creates an extra slash

cspeak:
Hello!

* Type: bug fix
* Link to issue:

Small description of change:

When generating models that contain more than one part to the namespace, the belongsTo relationship would add an extra slash in the Model reference.

Example: 
`phalcon all-models --get-set --doc --mapcolumn --namespace=App\\Models`

Any models that contains a "belongsTo" relationship is created as follows:
```php
$this->belongsTo('user_id', 'App\Models\\Users', 'user_id', ['alias' => 'Users']);
```

My patch will create the expected output:
```php
$this->belongsTo('user_id', 'App\Models\Users', 'user_id', ['alias' => 'Users']);
```

Since the "hasMany" relationships are created correctly, I just copied the namespace code from hasMany and applied it to belongsTo. I also tested and confirmed this works with larger namespaces such as "--namespace=My\\App\\V1\\Models"

Thanks


sergeyklay:
Could you please submit just the commits that belong to the PR?

PR #1151, opened 2017-11-19
Phalcon\Config from array

betrixed:
Hello!

* Type: bug fix
* Link to issue:

- [X] I have read and understood the [Contributing Guidelines][:contrib:]
- [X] I have checked that another pull request for this purpose does not exist
- [X] I wrote some tests for this PR

Small description of change:

In `scripts/Phalcon/Scanners/Config.php` if the` $probablyConfig` is an array, then it needs to become specifically a `Phalcon\Config` object, as an instance is expected by the calling load method.

`new Config($probableConfig)` returns the wrong kind of `Config` object. (instance of `$this`) should be new `PhConfig($probableConfig)`

Reason for encountering this: --  I have wedded the devtools / webtools to use my sites `config/config.php` which happens to return an array.  Of course I'm having a doubt about having a webtools compatible configuration in the middle of my differently arranged settings, but while it is so, this makes the devtools worthwhile for me.

Thanks

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -3,3 +3,8 @@ parameters:
 
     excludes_analyse:
         - %rootDir%/../../../scripts/Phalcon/Generator/*
+        - %rootDir%/../../../scripts/Phalcon/Commands/DotPhalconMissingException.php
+    ignoreErrors:
+        - '#Call to an undefined method object::toArray().#'
+        - '#supplied for foreach, only iterables are supported#'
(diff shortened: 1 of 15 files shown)


sergeyklay:
@betrixed  We don't add changes directly to stable branches or `master`. Could you please send this PR to the `3.2.x` branch?

sergeyklay:
@betrixed Could you please split this PR into two new:

* Bug fixes
* New feature request

Also note we don't add changes directly to stable branches or `master`. You have to send PR to the `3.2.x` branch.

Thank you for contributing

PR #1160, opened 2017-12-12
Corrected the Parentheses in 'if' on line 598

amsharma9:
Hello!

Small description of change:
The correct call to file_exists() should be with $viewPath as parameter AFAIK. The call should be file_exists($viewPath). This could be a potential bug.

Thanks


sergeyklay:
Thank you

Issue #1165, opened 2017-12-18
Make provision to skip fields when generating Views


amsharma9:
NFR: I wanted to know if we can add a provision to skip fields specified in `config.ini` file while generating `Views` like this:
```
fieldsToSkip   = created_by, creation_date, modified_by, modified_date
```
Developers might not want these fields to be placed in generated `views` as they are mostly updated in the background. If this feature is acceptable to DevTools team then I already have some code written which I can commit.

PR #1166, opened 2017-12-19
Make provision to skip fields when generating Views

amsharma9:
Changes as specified in Issue #1165.

Kindly review if this code adheres to your standards and that it is correct and efficient.

Thanks

### DIFF
--- a/scripts/Phalcon/Builder/Scaffold.php
+++ b/scripts/Phalcon/Builder/Scaffold.php
@@ -35,6 +35,8 @@ use Phalcon\Db\Column;
  */
 class Scaffold extends Component
 {
+	private $fieldsToSkip = array();
+    
     /**
      * @param string $fieldName
      *
@@ -91,6 +93,9 @@ class Scaffold extends Component
         $name = $this->options->get('name');
         $config = $this->getConfig();
 
+		if (isset($config->application->fieldsToSkip)) {		
+			$this->fieldsToSkip = array_map('trim', explode(",", $config->application->fieldsToSkip));
+		}
         if (empty($config->path('database.adapter'))) {
             throw new BuilderException('Adapter was not found in the config. Please specify a config variable [database][adapter].');
         }
@@ -426,6 +431,8 @@ class Scaffold extends Component
 
         $code = '';
         foreach ($this->options->get('dataTypes') as $attribute => $dataType) {
+			if (in_array($attribute, $this->fieldsToSkip))
+				continue;
             if (($action == 'new' || $action == 'edit') && $attribute == $identityField) {
                 continue;
             }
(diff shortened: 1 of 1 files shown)


amsharma9:
There is some issue in integration. I read the log but didn't understand the exact reason for the failure. Kindly help.

sergeysviridenko:
@amsharma9 I'll try to sort out with this asap.

sergeysviridenko:
@amsharma9 I've took a look your PR and issue. Why do you apply this only for scaffold? If we add option to config, users expect this behavior in all project. I would expected this in model, for example. To add option in config, that excluded db fields from generated files, only for scaffold isn't obviously and must be expanded for all project. 
About code. Please, separate your code in function with one action and call it where you need(if it possible).
Could you cover your code with small console test, please.

amsharma9:
@sergeysviridenko I think we should get it to work with Scaffold and then look at where else we can add it. Developers might only want the fields to be skipped in Views as the data still might need to be updated in those fields via models. That is what I am doing in my project.

I thought a lot about how to create a function to do this but looks non-intuitive to me. The lines for skipping have to be added where loops are running and generating code. If I create a function then even if we rewrite the while code we can't achieve this. There is only one place at which code is being generated and we need to write the skippiing code exactly at that place. Looks more or less impossible to separate out the skipping functionality.

amsharma9:
I have done testing of the code. What would you like me to show you. Kindly LMK how do I pass on the test and its result to you.

sergeysviridenko:
@amsharma9 
About add new features - I'm sure if I add `fieldsToSkip` - I expect fields will be skipped in scaffold, model etc. Why we add it to scaffold, but don't add to other places where it can be implemented. I think feature must be added everywhere or don't add anywhere.

About testing - I believe, it works, but we write tests for the future too. Code changes all time and all features must testing all time. So we try to cover code by tests.
Also all conflict must be solved.

amsharma9:
Ok, I understand your point of view. I suggest we create two options `fieldsToSkipInView' and 'fieldsToSkipInModel`. Does that look like a good idea to you.

sergeysviridenko:
@amsharma9 yes you're right, this is good way. First at all this option must be added to all places where possible. Repeated code should be separated to method. And all changes should be covered by tests. Finally, make PR to 3.3.x branch, please.
Thank you very much.

amsharma9:
@sergeysviridenko I checked the code of Model generation, it already has an `excludefields` option to probably do what we are trying to do using `skipFieldsInView`. See line 346 onwards. Can you possibly check. If this is doing exactly the same then we can change our option to `excludeFieldsInView` to maintain consistency of syntax.

--excludefields=l
  | Excludes fields defined in a comma separated list [optional]

sergeysviridenko:
@amsharma9 ok thank you. I'll take a look asap.

PR #1255, opened 2018-10-23
Add option to ignore current schema on initialize function

huydang284:
Hello!

* Type: New feature
* Link to issue: None

**In raising this pull request, I confirm the following (please check boxes):**

- [x] I have read and understood the [Contributing Guidelines][:contrib:]
- [x] I have checked that another pull request for this purpose does not exist
- [ ] I wrote some tests for this PR

Small description of change: I want to use generated model in different servers (different schemas) so I don't want to add schema to model's initialize function.

Thanks

[:contrib:]: https://github.com/phalcon/phalcon-devtools/blob/master/CONTRIBUTING.md

Issue #1162, opened 2017-12-14
Hardcoding schema in init


amsharma9:
DevTools hard codes `schema` in `initialize()` function of generated models. Is this a good idea? Hard coding database name into the code creates so many issues and reduces flexibility. Does it serve any purpose? See this query [Phalcon tools setSchema by default now?](https://forum.phalconphp.com/discussion/15823/phalcon-tools-setschema-by-default-now) on the forum.

Is there a way to avoid this.

sergeyklay:
@amsharma9 Could you please provide typical not neatly generated initialize method?

amsharma9:
Here is a sample of code generated:

```
    /**
     * Initialize method for model.
     */
    public function initialize()
    {
        $this->setSchema("testdb");
        $this->setSource("co_bank_m");
    }
```

sergeyklay:
@amsharma9 The last thing I need - steps to reproduce :)

amsharma9:
Normal DevTools command for model generation. It generates this code with or without any options AFAIK.
```
phalcon model --name co_bank_m --get-set --namespace=Baseapp\Models --mapcolumn --extends=ModelBase
```

sergeysviridenko:
@amsharma9 If I understood correctly, you would like to create model without `$this->setSchema("testdb");` in `initialize`. you can add option `--schema=''`.

Could you try run 
```
phalcon model --name co_bank_m --get-set --namespace=Baseapp\Models --mapcolumn --extends=ModelBase --schema=''
```

amsharma9:
@sergeysviridenko It didn't work. It is looking for the table `co_bank_m` in blank database and it gives error:

```
 Error: Table "co_bank_m" does not exist.
```
If I change `schema` to my DB name `testdb` then it goes through.

sergeysviridenko:
@amsharma9 Look I have table `member`. I've run `phalcon model --name member --get-set --namespace=Baseapp\Models --mapcolumn --extends=ModelBase --schema=''` and all works fine. Model has been created without `$this->setSchema("testdb");`.
Could you try to use this command with created table.

amsharma9:
Ok, what you wrote works but I had to remove the empty quotes. With single quotes after schema it still gave `ERROR: Table "co_bank_m" does not exist.`. I am on Windows 7 using Wamp so could that be a reason for it not to work. This is what I gave on command line when it worked:
```
[...]

PR #1304, opened 2019-05-21
fix: should not setSchema('foo') in initialize when we set --schema=""

dacgray:
Hello!

* Type: bug fix
* issue:

#1162
This may be a better solution than #1255 

We expect that setting --schema="" will not add `setSchema()` in `initialize()`.

**Please check if fix provides desired behaviour.  If not I'll look into fixing #1255  ^^**

---

- [x] I wrote some tests for this PR

---

- another pull request for this issue exists, but looks stale and this may be the better solution.
- this PR passes the option schema value from `AllModels::build` to `Models`.  
- we still use the schema from config to get our adapter instance
- this allows behaviour as described here: https://forum.phalconphp.com/discussion/15823/phalcon-tools-setschema-by-default-now

Thanks

### DIFF
--- a/scripts/Phalcon/Builder/Model.php
+++ b/scripts/Phalcon/Builder/Model.php
@@ -94,6 +94,48 @@ class Model extends Component
     }
 
     /**
+     * @return ModelOption
+     */
+    public function getModelOptions()
+    {
+        return $this->modelOptions;
+    }
+
+    /**
+     * We should expect schema to be string|null
+     * OptionsAware throws when getting null option values
+     * so we need to handle shouldInitSchema logic with the raw $option array
+     *
+     * Should setSchema in initialize() only if:
+     * - $option['schema'] !== ''
+     *
+     * @return bool
+     */
+    public function shouldInitSchema()
+    {
+        return !isset($this->modelOptions->getOptions()['schema'])
+            || $this->modelOptions->getOptions()['schema'] !== '';
+    }
+
+    /**
+     * @return string
+     */
+    public function getSchema()
+    {
+        if ($this->modelOptions->hasOption('schema') && !empty($this->modelOptions->getOption('schema'))) {
+            $schema = $this->modelOptions->getOption('schema');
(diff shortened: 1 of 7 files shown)


scripts/Phalcon/Builder/Model.php
```diff
@@ -159,7 +159,7 @@ public function build()
             $schema = Utils::resolveDbSchema($config->database);
         }
 
-        if ($schema) {
+        if (!empty($this->modelOptions->getOption('schema'))) {
```
Jeckerson: There is same check on line 156

```
if ($this->modelOptions->hasOption('schema'))
```

dacgray: `hasOption` uses isset.

The schema option is always set when coming from AllModels, so `$this->modelOptions->hasOption('schema')`, so we always setSchema in initialize.

If we want to follow the logic described here: https://forum.phalconphp.com/discussion/15823/phalcon-tools-setschema-by-default-now

We need to check if !empty.

If we don't want to follow the logic described in the forum post we need an init-schema flag.

dacgray:
@Jeckerson 

What do you say?

- move forward with this solution 
- or fix https://github.com/phalcon/phalcon-devtools/pull/1255

Jeckerson:
I'll check and tell you as soon as possible.

Jeckerson:
@dacgray Can you create separate tests without altering current? To make sure that script executing that line `$schema = Utils::resolveDbSchema($config->database);`

dacgray:
@Jeckerson @sergeyklay please take another look.

dacgray:
@Jeckerson @sergeyklay good to merge?

dacgray:
@Jeckerson @sergeyklay is there anything I need to do, or do you have a timeline for a 3.4.2 release?

Jeckerson:
@dacgray https://github.com/phalcon/phalcon-devtools/milestone/4

Issue #1378, opened 2019-11-23
Support global usage


ruudboon:
When installing devtools (v4.0.0-beta.2) globally you will get the following error:
```
ERROR: Please run composer install
```
Please add support for global installation.

sergeyklay:
Do you mean [phalcon.phar](https://github.com/phalcon/phalcon-devtools/releases/download/v4.0.0-beta.2/phalcon.phar)?

Jeckerson:
@sergeyklay No, when you install package globally: `composer global install`.

faridanthony:
Hi, is this solved yet? I'm getting the same:
ERROR: Please run composer install

foroughfadaei:
> Hi, is this solved yet? I'm getting the same:
> ERROR: Please run composer install

me too

mrdipesh1:
Samething happened to me for 2 days, but finally it is solved.
I did,

1. Installed composer, by downloading composer.exe

2. Open composer.json which is located at C:/user/Your PC USERNAME/appdata/roaming/composer and pasted
{
"require-dev": {
"phalcon/devtools": "~4.0"
}
}

3. run `composer global require phalcon/devtool` in cmd. It will take time to download, you have to wait. 
Note: I have installed GIT, [From here](https://www.atlassian.com/git/tutorials/install-git) and during the process it will ask for token key and the link for that will be given by cmd itself. If any error occured during download, restart this step. Phalcon/devtool must be @100% or should be installed through cache. I didnt care about other downloads excepts downloads with name phalcon/(migration & devtools).

4. After complete, you will find the phalcon devtool path inside that composer path

5. inside devtool path, there is a file in .bat format. Open it and modify the PTOOLPATH to same directory where that .bat file located

6. register the same path you just used in .bat file in Enviromental Variable PATH

7. open new CMD and type phalcon. I got the result as it should display.

fagai:
Will this support not be merged into 3.x?

Jeckerson:
Related #1431

DebugTheCode:
Going into the phalcon-devtools folder (e.g. /Users/[username]/phalcon-devtools and run `composer install`, fixed it for me.

javad4481:
run this in terminal
below in [username] insert your own username  like   "/home/javad/phalcon-devtools"
`cd /home/[username]/phalcon-devtools/`
`composer install`
if " phalcon " command not found
`alias phalcon=/home/[username]/phalcon-devtools/phalcon`

[...]

PR #1379, opened 2019-11-26
Import global vendor of Composer

jenovateurs:
Hello!

* Type: bug fix
* Link to issue:  #1378 

Small description of change:
Check new path for autoload. This path is useful when you install phalcon-devtools with composer in global or local way.
If the path exists, I update the variable $vendorAutoload to load the correct file.

Thanks

### DIFF
--- a/bootstrap/autoload.php
+++ b/bootstrap/autoload.php
@@ -99,15 +99,21 @@ defined('COMPATIBLE_VERSION') || define('COMPATIBLE_VERSION', 3020040);
 /**
  * Register the Composer autoloader (if any)
  */
-$vendorAutoload = PTOOLSPATH . DS . 'vendor' . DS . 'autoload.php';
-if (!file_exists($vendorAutoload)) {
-    $vendorAutoload = PTOOLSPATH . DS . '..' . DS . '..' . DS . 'autoload.php';
-    if (!file_exists($vendorAutoload)) {
-        throw new Exception('Please run composer install');
+$vendorAutoload = [
+    PTOOLSPATH . DS . 'vendor' . DS . 'autoload.php', // Is installed locally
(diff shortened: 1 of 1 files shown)


bootstrap/autoload.php
```diff
@@ -101,7 +101,10 @@
  */
 $vendorAutoload = PTOOLSPATH . DS . 'vendor' . DS . 'autoload.php';
 if (!file_exists($vendorAutoload)) {
-    throw new Exception('Please run composer install');
+    $vendorAutoload = PTOOLSPATH . DS . '..' . DS . '..' . DS . 'autoload.php';
```
Jeckerson: In what directory you are trying to find autoload.php ?

jenovateurs: I installed phalcon-devtools with `composer global require phalcon/devtools:~v4.0.0-beta.2`.
Then I tried to use phalcon and an error appeared : 
`ERROR: Please run composer install`
I searched in the phalcon-devtools code and I saw the content of the `$vendorAutoload` variable.
In Debian, the value is  `.composer/vendor/phalcon/devtools/vendor/autoload.php`.
In Windows 10, the value is `\AppData\Roaming\Composer\vendor\phalcon\devtools\vendor\autoload.php`.
The right `autoload.php` file is in `.composer/vendor/autoload.php` or `\AppData\Roaming\Composer\vendor\autoload.php`. That's why I'm offering this solution.
Tested on Windows 10 and Debian Stretch with Docker.

sergeyklay: We need something like this https://github.com/phalcon/zephir/blob/development/Library/autoload.php

jenovateurs: You're right, it's easier to read and understand.

Issue #1412, opened 2020-01-03
webtools don't work (404)


zak-wojtek77:
### Expected and Actual Behavior
After run command:
vendor/bin/phalcon webtools enable
and run webside:
http://myWebside/webtools.php

I get 404

### Details

* System info and versions (if possible): Phalcon-devtools 4.0.0
* Phalcon Framework version: 4.0.0
* PHP Version: 7.4
* Operating System: Linux Debian
* Server: Nginx
* Other related info (Database, table schema): Mysql 8

Jeckerson:
Known bug, use `/` at the end, like that: `webtools.php/`

zak-wojtek77:
Really now if I add at the end "/" I can see website. But still don't know how I can generate something. This problem was not in Phalcon 3.4.1.

Jeckerson:
Now you can't use both: `notFound()` and `setDefault*()` in Router. If you declare `notFound()`, then any routes that are not in the list - will send to you 404 action.

Jeckerson:
Fixed in #1413

PR #1416, opened 2020-01-12
4.0.1

Jeckerson:
Next release

* Fixed WebTools root `/` uri access (#1412)


tests/acceptance/Web/Tools/Controllers/ErrorControllerCest.php
```diff
@@ -11,8 +11,10 @@ final class ErrorControllerCest
      * @covers \Phalcon\Devtools\Web\Tools\Controllers\ErrorController::route404Action
      * @param AcceptanceTester $I
      */
-    public function testRoute404Action(AcceptanceTester $I): void
+    public function testRoute404Action(AcceptanceTester $I, $scenario): void
     {
+        $scenario->skip('Deprecated approach');
```
niden: For the future, you can create a method in your AcceptanceTester which will throw the necessary exception so you can use `$I->skipTest()` here.

Issue #1438, opened 2020-03-29
[BUG]: Phalcon WebTools wrong migration directory


wajdijurry:
Hello,

I am using Phalcon WebTools to manage my app staff, it reads from "config.php" file where I specified all my app configs like modelsDir, controllersDir and migrationsDir, this is a part of my config.php file:
```
    'application' => [
        'modelsDir' => APP_PATH . '/common/models/',
        'controllersDir' => APP_PATH . '/modules/api/controllers/',
        'migrationsDir' => APP_PATH . '/migrations/',
        'logsDir' => APP_PATH . '/logs/',
        'uploadDir' => BASE_PATH . '/public/uploads/'
    ],
```
so when I came to Migrations view to create a new migration, it creates the migration file in a wrong directory, under root dir (/1.0.0), while it is supposed to create it under (/var/www/html/app/migrations) as specified in the form:
![Screenshot from 2020-03-29 20-32-33](https://user-images.githubusercontent.com/25797257/77855940-8842d500-71fc-11ea-9acf-3f22d23f73e8.png)

After digging into the code, I found a bug in Migrations script, which tries to read the migrations directory as array at first ```count($migrationsDirs) > 1```, then if not an array, it takes the first character of the string as the root path (else block, ```$migrationsDir = $migrationsDirs[0]```), so the result of "/var/www/html/app/migrations/" will be "/" which is wrong.
```
        if (count($migrationsDirs) > 1) {
            $question = 'Which migrations path would you like to use?' . PHP_EOL;
            foreach ($migrationsDirs as $id => $dir) {
                $question .= " [{$id}] $dir" . PHP_EOL;
            }
            fwrite(STDOUT, Color::info($question));
            $handle = fopen("php://stdin", "r");
            $line = (int)fgets($handle);
            if (!isset($migrationsDirs[$line])) {
                echo "ABORTING!\n";
                return false;
            }
            fclose($handle);
            $migrationsDir = $migrationsDirs[$line];
        } else {
            $migrationsDir = $migrationsDirs[0];
        }
```
### Expected and Actual Behavior

Expected:
```
/var/www/html/app/migrations/
```

Actual:
```
/
```

### Details

* System info and versions (if possible): (`phalcon info`)
* Phalcon Framework version: (`php --ri phalcon`)
phalcon

Web framework delivered as a C-extension for PHP
phalcon => enabled
Author => Phalcon Team and contributors
Version => 3.4.4
[...]

PR #1439, opened 2020-03-29
Fix Bug #1438

wajdijurry:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/phalcon-devtools/issues/1438

Small description of change:
Migrations script now can handle the migrations directory well.

Thanks

### DIFF
--- a/scripts/Phalcon/Migrations.php
+++ b/scripts/Phalcon/Migrations.php
@@ -87,24 +87,26 @@ class Migrations
         $optionStack->setDefaultOption('noAutoIncrement', null);
         $optionStack->setDefaultOption('verbose', false);
 
-        $migrationsDirs = $optionStack->getOption('migrationsDir');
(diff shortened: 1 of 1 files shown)


scripts/Phalcon/Migrations.php
```diff
@@ -87,24 +87,26 @@ public static function generate(array $options)
...
+                $line = (int)fgets($handle);
+                if (!isset($migrationsDirs[$line])) {
+                    echo "ABORTING!\n";
+                    return false;
+                }
+                fclose($handle);
+                $migrationsDir = $migrationsDirs[$line];
+            } else {
+                $migrationsDir = $migrationsDirs[0];
             }
```
Jeckerson: You forgot to add here else, in case if `migrationsDir` is a string.

wajdijurry: This is already handled in line 90, please check

PR #1476, opened 2020-11-16
Generated models::findFirst doesnt have a return type

talal424:
Hello!

* Type: bug fix

Small description of change:

when generating a new model the ```findFirst``` method doesn't have a return type and results in a fatal error

>PHP Fatal error:  Declaration of TestModel::findFirst($parameters = NULL) must be compatible with Phalcon\Mvc\Model::findFirst($parameters = NULL): ?Phalcon\Mvc\ModelInterface

fixed with adding return type compatible with Phalcon\Mvc\Model::findFirst

Thanks

### DIFF
--- a/src/Generator/Snippet.php
+++ b/src/Generator/Snippet.php
@@ -265,9 +265,9 @@ EOD;
      * Allows to query the first record that match the specified conditions
      *
      * @param mixed \$parameters
-     * @return %s|\Phalcon\Mvc\Model\ResultInterface
+     * @return %s|null
      */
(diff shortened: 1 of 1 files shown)


Jeckerson:
This won't be correct, as `v4.0` has a bug and returns `false` instead of `null`.

talal424:
yeah i didn't know that

i'll convert this to a draft or delete it

Jeckerson:
No problem. I'll close. In case you have any another code suggestions - feel free to submit PR.

Issue #1317, opened 2019-09-14
Class constants vanish when model is regenerated with --force


lmfmaier:
Phalcon DevTools Version: 3.4.0
Phalcon Version: 3.4.4

After regenerating a model with `phalcon model <model> --force` class constants defined at the top of the class definition are gone. Custom class methods and members defined before the regeneration are still there

BeMySlaveDarlin:
Fixed this bug and some other bugs related to `--force`

PR #1498, opened 2021-03-22
ISSUE-1317: Fix model --force creation bugs

BeMySlaveDarlin:
Hello!

* Type: bug fix
* Link to issue: #1317 

Small description of change:
Fixed constants dropping in models when recreating with `--force` option: constants with access modifiers (public, protected, private) were ignored by preg_match pattern.
Fixed attributes duplication bug when recreating with `--force` option if no `--camelize` option declared and property is snake_cased.
Fixed column mapping on creation from table with PascalCase fields.

Thanks

### DIFF
--- a/src/Builder/Component/Model.php
+++ b/src/Builder/Component/Model.php
@@ -27,6 +27,8 @@ use Phalcon\Text;
 use Phalcon\Validation;
 use Phalcon\Validation\Validator\Email as EmailValidator;
 use ReflectionClass;
+use ReflectionClassConstant;
+use ReflectionProperty;
 
 /**
  * Builder to generate models
@@ -121,13 +123,8 @@ class Model extends AbstractComponent
         $genDocMethods = $this->modelOptions->getValidOptionOrDefault('genDocMethods', false);
         $useSettersGetters = $this->modelOptions->getValidOptionOrDefault('genSettersGetters', false);
 
-        $adapter = $config->database->adapter;
+        $adapter = $config->database->adapter ?? 'Mysql';
         $this->isSupportedAdapter($adapter);
-
-        $adapter = 'Mysql';
-        if (isset($config->database->adapter)) {
-            $adapter = $config->database->adapter;
-        }
 
         if (is_object($config->database)) {
             $configArray = $config->database->toArray();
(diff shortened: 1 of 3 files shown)


src/Builder/Component/Model.php
```diff
@@ -571,6 +498,65 @@ public function build(): void
...
+    }
+
+    /**
+     * @param string $fieldName
+     *
+     * @return string
+     */
+    protected function getFieldName($fieldName)
+    {
+        if ($this->modelOptions->getOption('camelize')) {
```
Jeckerson, line 553:
```suggestion
        if ($this->modelOptions->hasOption('camelize')) {
```

Probable `hasOption` is more correct here?

src/Builder/Component/Model.php
```diff
@@ -571,6 +498,65 @@ public function build(): void
...
+
+        return null;
+    }
+
+    /**
+     * @param string $fieldName
+     *
+     * @return string
+     */
+    protected function getFieldName($fieldName)
```
Jeckerson, line 551:
```suggestion
    protected function getFieldName(string $fieldName)
```

src/Builder/Component/Model.php
```diff
@@ -571,6 +498,65 @@ public function build(): void
...
     }
 
+    /**
+     * @param array $linesCode
+     * @param string $pattern
+     * @param ReflectionProperty|ReflectionClassConstant $attribute
+     *
+     * @return null|string
+     */
+    protected function getAttribute($linesCode, $pattern, $attribute)
```
Jeckerson, line 508:
```suggestion
    protected function getAttribute(array $linesCode, string $pattern, $attribute): ?string
```

Jeckerson (changes requested):

src/Builder/Component/Model.php
BeMySlaveDarlin: `camelize` is set during initialization to boolean value (`true` if set via command or `false` by default).
`hasOption` will return always true is this case

BeMySlaveDarlin: More correct check could be
```
if ($this->modelOptions->hasOption('camelize') && $this->modelOptions->getOption('camelize')) {
```
But seems overhead

Jeckerson: Ok

Jeckerson (approved):

PR #1530, opened 2022-10-21
Fix missing return type for _getVersion()

Bellardia:
We're still using 3.4.x on PHP 7.4

There's a type error caused by a missing return type that was added to Phalcon\Version.

```
Fatal error: Declaration of Phalcon\Devtools\Version::_getVersion() must be compatible with Phalcon\Version::_getVersion(): array in /usr/share/phalcon-devtools/scripts/Phalcon/Devtools/Version.php on line 40
```

### DIFF
--- a/scripts/Phalcon/Devtools/Version.php
+++ b/scripts/Phalcon/Devtools/Version.php
@@ -37,7 +37,7 @@ class Version extends PhVersion
      *
      * @return array
      */
-    protected static function _getVersion()
+    protected static function _getVersion(): array
     {
         return [3, 4, 14, 0, 0];
     }


rudiservo (commented):
Phalcon 3.4 no longer maintainable.

Bellardia:
> Phalcon 3.4 no longer maintainable.

It kind of is; we're running it in production on PHP 8.2. Whether or not you want these changes merged back is a choice from there. I have a much larger PR for full compatibility that there seemed to be interest for in discord, even if it only benefits a select few.

rudiservo:
Why have 3.4 in php8.2 and why haven't you upgraded yet to 5.

There are other issues why I can't accept this, return types where only after php7.4
Phalcon 3.4.4 is can be used in php 7, although php7 is EOL, this means not only I would break compatibility, I would have to be responsible for maintain this.

https://www.php.net/manual/en/language.types.declarations.php

I hope you understand but time is limited, I would rather focus phalcon 5 for now, if you need something to make your upgrade easier just create an issue.

Bellardia:
All good, I actually totally understand, feel free to close.
We face opposite constraints :) In this case, it's far easier to maintain 3.4.x on modern PHP than it is for me to upgrade to Phalcon 5. It'd be great to move to 5, but the changes are largely cosmetic so there hasn't been a business case.

rudiservo:
Current Phalcon5 is not just cosmetic, there are bug fixes, try it and let me know.
Thanks for the contribution anyway :)

PR #1534, opened 2023-03-10
v5 version

lmfmaier:
Changes/migrations according to https://docs.phalcon.io/5.0/de-de/upgrade

- Change namespaces in code and templates
- Remove usages of Phalcon\Text, replace them by Phalcon\Support\Helper
- Remove Phalcon\Exception, use \Exception instead
- Replace registerDirs() with setDirectories() in (autoloader) in templates


lmfmaier:
Succesfully created a project and generated a model class from local database table. Successfully ran the generated project skeleton ("You're now flying with Phalcon. Great things are about to happen!") on PHP 8.1 with Phalcon 5.1

